PR #47, opened 2014-03-02
[admintools] install useful packages for system and network administrators

jremond:
Just an idea to ensure some tools are present for admins.
My test environment is CentOS6.


peckpeck:
We don't include techniques directly in ncf. This however a good example.

PR #58, opened 2014-05-20
Refs #4873 - Add bundle command_execution_module() and adapt command_execution() and its tests

nperron:
Refs #4873 - Add bundle command_execution_module() and adapt command_execution() and its tests

cf http://www.rudder-project.org/redmine/issues/4873


peckpeck:
We have no current use case for this.

PR #87, opened 2014-07-15
pull request mounting nfs by array

Webhuis:
mount_nfs_array is the abstrated policy.
mnt_common_backups is an example on how to use it.


peckpeck:
This is not generic enough, see 
https://www.rudder-project.org/redmine/issues/10900

PR #93, opened 2014-08-08
Fixes #5378: We need a method for the mustache template

peckpeck:
Copied from file_template_expand.cf


peckpeck:
PR rebased

amousset:
Wrong one!

peckpeck:
yes, just a failed test with rudder-dev

PR #123, opened 2014-09-17
Fixes #5550: correct classes conditions when using quotes

ncharles:


tools/ncf_rudder.py
```diff
@@ -271,7 +273,7 @@ def generate_rudder_reporting(technique):
     method_name = method_call['method_name']
     generic_method = generic_methods[method_name]
 
-    key_value = method_call["args"][generic_method["class_parameter_id"]-1]
+    key_value = method_call["args"][generic_method["class_parameter_id"]-1].replace("\\'", "\'").replace('\\"', '\"')
```
peckpeck: "\'" is equivalent to "'" in python

peckpeck: I don't understand this one, you added 2 more backslash before and here you remove only one.

ncharles: i guess this is not relevant anymore; we changed a lot of things around that (notably, we don't use bash anymore to generate the reporting file)

peckpeck: If this is not relevant, should we remove some or all of this PR ?

ncharles:
actually, this is still relevant :(

ncharles:
replaced by https://github.com/Normation/ncf/pull/223

PR #136, opened 2014-10-22
Fixes #5675: Generate a caller bundle when creating a rudderified technique

peckpeck:
http://www.rudder-project.org/redmine/issues/5675


peckpeck:
PR updated

PR #137, opened 2014-10-22
Fixes #5679: Add technique variables support to ncf api

peckpeck:
http://www.rudder-project.org/redmine/issues/5679


peckpeck:
PR updated

PR #160, opened 2014-12-05
Fixes #5924: Clarify the README (quickstart, etc...)

Kegeruneku:
Ticket: https://www.rudder-project.org/redmine/issues/5924

To be merged in master

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,20 @@
+# API
 api/flask
 api/virtualenv.py
 local/*
+
+# Doc
+doc/all_generic_methods.txt
+doc/generic_methods.md
+doc/generic_methods.html
+
+# Site
+site/content/generic_methods.md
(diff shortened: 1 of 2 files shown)


ncharles:
This looks good to me !

jooooooon:
This edit seems perfect for GitHub's README. However, this README is also the main page on http://www.ncf.io... which I'm pretty sure we don't want to change like this (some changes are great - but removing syntax colouring in the main example, is not).

Could you please rework this so that these 2 pages are separate? It's probably a simple case of "cp README site/content/index.md" or similar.

Kegeruneku:
Updated according to review :)

VinceMacBuche:
@Kegeruneku, you should switch the readme and index.md, (ie the syntax highlight was removed from index.md and is still in readme)

amousset:
This has been done elsewhere, closing.

PR #167, opened 2015-01-16
Fixes #4806: distinguish the Technique we are doing report on

ncharles:


ncharles:
to be merged with https://github.com/Normation/cf-clerk/pull/67

peckpeck:
It's OK for me

PR #171, opened 2015-02-17
Fixes #6275: Fix negative status and test when a variable is used in class context

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6275


Kegeruneku:
Some extra unneeded concat() calls, but the overlall looks OK to me !

PR #172, opened 2015-02-18
Fixes #6281: rewrite wrong url preventing access to path variable

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6281


fanf:
Closing that one, see the ticket for more information (basically, the problem was not what we thought)

PR #173, opened 2015-02-19
Adding check for fakeroot.

danielhoherd:
Test for fakeroot, abort if not found.


Kegeruneku:
Thanks for the commit Daniel, merging!

PR #176, opened 2015-03-04
Fixes #5676: ncf builder doesn't cope well with rudder session timeout

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/5676


VinceMacBuche:
PR updated

PR #179, opened 2015-03-13
Fixes #6407: ncf-doc cannot parse generic methods new format

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6407


peckpeck:
OK Good!

PR #182, opened 2015-03-18
Fixes #6418: Add the file_download generic method

Kegeruneku:
Ticket: http://www.rudder-project.org/redmine/issues/6418

To be merged in master

### DIFF
--- a/tests/acceptance/30_generic_methods/network/file_download.cf
+++ b/tests/acceptance/30_generic_methods/network/file_download.cf
@@ -1,0 +1,83 @@
+#######################################################
+#
+# Test checking if:
+# * a file can be downloaded from a http
+# * a file that is already here will not be downloaded
+#
+#######################################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+
+    "tmp"                    string => getenv("TEMP", 1024);
+
+    "source_file"            string => "http://www.normation.com/robots.txt";
+    "destination_file"       string => "${tmp}/test_file_download.txt";
+    "destination_file_canon" string => canonify("${destination_file}");
+
+    "existing_file"          string => "${tmp}/test_file_download2.txt";
+    "existing_file_canon"    string => canonify("${existing_file}");
+
+
+  files:
+    "${existing_file}"
+      create => "true";
+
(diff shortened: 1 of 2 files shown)


tree/30_generic_methods/file_download.cf
```diff
@@ -0,0 +1,74 @@
...
+# This bundle will define a class file_download_${destination}_{kept,repaired,not_ok,ok,reached}
+
+bundle agent file_download(source, destination)
+{
+
+  vars:
+
+    curl_utility_present::
+
+      "action_command"            string => "${unix_knowledge.utility[curl]} -s -L -o ${destination} ${source}";
```
ncharles: you are deliberatly making it not compatible with non_unix ?

tree/30_generic_methods/file_download.cf
```diff
@@ -0,0 +1,74 @@
...
+
+      "report"                    usebundle => _logger("Download ${source} into ${destination}", "${class_prefix}"),
+                                 ifvarclass => "${class_prefix}_reached";
+
+    !curl_utility_present.!wget_utility_present.!file_exists::
+
+      "force_failure_class"       usebundle => _classes_failure("${class_prefix}");
+      "report"                    usebundle => _logger("Unable to download ${source}: neither wget or curl are installed", "${class_prefix}");
+
+    file_exists::
```
ncharles: are you sure this doe snot get evaluated at the second run as a success ?

tests/acceptance/30_generic_methods/network/file_download.cf
```diff
@@ -0,0 +1,57 @@
...
+#######################################################
+
+bundle agent init
+{
+  vars:
+
+    "tmp"                    string => getenv("TEMP", 1024);
+
+    "source_file"            string => "http://www.normation.com/robots.txt";
+    "destination_file"       string => "${tmp}/test_file_download.txt";
```
ncharles: I would also like a test where the file is already there, so don't need to download it

tree/30_generic_methods/file_download.cf
Kegeruneku: For now yes, as there is no "standard" utility on Windows to download files like wget or curl. It can always be added later.

tree/30_generic_methods/file_download.cf
Kegeruneku: Updated the class conditions so that it never happens.

tests/acceptance/30_generic_methods/network/file_download.cf
Kegeruneku: Done

Kegeruneku:
PR updated according to review.

tree/30_generic_methods/file_download.cf
ncharles: wait, we do have this defined in paths already
you could simply use ${paths.path[curl]} and ${paths.path[wget]}

Kegeruneku:
Updated again using paths.path[XXX]

ncharles:
Awesome !
Thank you

PR #184, opened 2015-03-27
Fixes #6440: We can't use é in the generic methods parameters

peckpeck:
http://www.rudder-project.org/redmine/issues/6440

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -15,6 +15,7 @@ import os.path
 import shutil
 import sys
 import os
+import codecs
 
 # Additionnal path to look for cf-promises
 additional_path = ["/opt/rudder/bin","/usr/sbin","/usr/local"]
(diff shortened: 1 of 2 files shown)


peckpeck:
PR updated

tools/ncf.py
```diff
@@ -147,7 +146,7 @@ def parse_bundlefile_metadata(content, bundle_type):
         else:
           res[tag] = match.group(1)
 
-    match = re.match("[^#]*bundle\s+agent\s+(\w+)(\(([^)]+)\))?.*$", line)
+    match = re.match("[^#]*bundle\s+agent\s+([a-zA-Z_]+)(\(([^)]+)\))?.*$", line, flags=re.UNICODE)
```
VinceMacBuche: It misses '0-9' in the regexp

VinceMacBuche:
@peckpeck Just this little change and you can self merge the PR

peckpeck:
PR updated

tools/ncf.py
```diff
@@ -147,7 +146,7 @@ def parse_bundlefile_metadata(content, bundle_type):
         else:
           res[tag] = match.group(1)
 
-    match = re.match("[^#]*bundle\s+agent\s+(\w+)(\(([^)]+)\))?.*$", line)
+    match = re.match("[^#]*bundle\s+agent\s+([0-9a-zA-Z_]+)(\(([^)]+)\))?.*$", line, flags=re.UNICODE)
```
ncharles: why not \w anymore ? is it to prévent accents in bundle names ? (this is legal in cfengine)

ncharles:
Oh, this change doesn't seem to work properly on Python 2.6 :(

PR #189, opened 2015-04-08
Fixes #6473: define job schedule on one line to fix tests

ncharles:


ncharles:
replaced by https://github.com/Normation/ncf/pull/191

PR #191, opened 2015-04-08
Fixes #6473: Job scheduler generic methods have multiline definition, which is not compatible with the test

peckpeck:
http://www.rudder-project.org/redmine/issues/6473


ncharles:
looks good, but can you change the commit name to be related to the change ?

peckpeck:
Done

PR #192, opened 2015-04-17
use long option names in Quickstart

thomas-koch-comsolit:
When I write documentation I always use the long format of options. Even in shell scripts I prefer the long option since I might come back and have to read the script.

I use the short format of options only when I directly work in the shell.


peckpeck:
@thomas-koch-comsolit 
You're right about long options!
Merging.
Thank you very much.

PR #193, opened 2015-04-27
Fixes #6533: Move ncf examples out of the main tree to prevent errors in...

Kegeruneku:
... ncf-builder

Ticket: http://www.rudder-project.org/redmine/issues/6533

To be merged in master


VinceMacBuche:
I think the services and the knowledge should be put in examples too!

Kegeruneku:
Not sure about that... Don't we reference them in quickstarts / demos ?

VinceMacBuche:
We should update demo to tell to use examples it would be better

VinceMacBuche:
Can you open an issue about that, removing techniques is more important that having good services/knowledge.

VinceMacBuche:
So i can merge this and we still track down that issue

Kegeruneku:
PR updated.

PR #194, opened 2015-04-28
Fixes #6539: test broken since normation.com has changed

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6539


peckpeck:
OK good

PR #195, opened 2015-04-28
Fixes #6534: Displaying the technique list in ncf-builder can take more than one minute

peckpeck:
http://www.rudder-project.org/redmine/issues/6534


peckpeck:
@VinceMacBuche ping

VinceMacBuche:
Awesome , thanks @peckpeck  ;)

ncharles:
we need to mimic that for the cf-promises used in promises generation !

PR #196, opened 2015-04-29
Fixes #6542: url for non existant test is wrong

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6542


Kegeruneku:
OK !

PR #198, opened 2015-05-12
Fixes #6531: accept to save technique on systems with Python2.6

ncharles:


peckpeck:
Ok good
Ticket: http://www.rudder-project.org/redmine/issues/6531

PR #199, opened 2015-05-19
Fixes #6615: Cannot save technique with ncf builder

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6615


ncharles:
great !

peckpeck:
OK good

PR #200, opened 2015-05-20
Fixes #6621: The file_ensure_line_present_in_ini_section bundle does not handle line editing

amousset:
http://www.rudder-project.org/redmine/issues/6621

### DIFF
--- a/api/ncf_api_flask_app/views.py
+++ b/api/ncf_api_flask_app/views.py
@@ -39,13 +39,20 @@ def check_authentication_from_rudder(auth_request):
     acl = "read"
   else:
     acl = "write"
-  # We skip ssl certificate verification, since rudder and ncf-api are on the same domain and same virtualhost
-  auth_result = requests.get('https://localhost/rudder/authentication?acl=' + acl, cookies =  request.cookies, verify = False)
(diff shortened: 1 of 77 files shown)


tree/30_generic_methods/file_ensure_line_present_in_ini_section.cf
```diff
@@ -20,43 +20,30 @@
...
+# @parameter name Name of the key to add or edit
+# @parameter value Value of the key to add or edit
 #
 # @class_prefix file_ensure_line_present_in_ini_section
 # @class_parameter file
 #
 # This bundle will define a class file_ensure_line_present_in_ini_section_${file}_{kept,repaired,not_ok,ok,not_ok,reached}
 
-bundle agent file_ensure_line_present_in_ini_section(file, section, line)
+bundle agent file_ensure_line_present_in_ini_section(file, section, name, value)
```
ncharles: i'm not sure about changing the signature - this may break user existing code

ncharles: this should be:
file_ensure_keys_values_in_ini_section

ncharles:
@amousset could you retarget this pr to branch V0.x ?
Thank you !

PR #201, opened 2015-05-20
Fixes #6622: Non-breaking spaces in comments in ncf files

amousset:
http://www.rudder-project.org/redmine/issues/6622


ncharles:
we would greatly benefit from a unit test that checks if there is invalid spaces in ncf

amousset:
There is already a test for this but it excludes comments, so maybe there is a reason ?

peckpeck:
There is a reason : there was too many changes to do for something that was not breaking anything.
But now that the work is done ...

PR #206, opened 2015-05-27
Fix acceptance test operation on AIX/HPUX/etc

rootwyrm:
AIX/HPUX/etc do not accept -printf args to find. Removing it allows acceptance tests to run on these platforms.


peckpeck:
We now use list-compatible-inupt instead of a find. This PR is not useful anymore

PR #208, opened 2015-06-22
Fixes #6436: Adding a carriage return in the description of a technique breaks the generated technique

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6436

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -30,6 +30,8 @@ tags["common"] = ["bundle_name", "bundle_args"]
 tags["generic_method"] = ["name", "description", "parameter", "class_prefix", "class_parameter", "class_parameter_id"]
 tags["technique"] = ["name", "description", "version"]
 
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
PR updated

tools/ncf.py
```diff
@@ -417,8 +435,10 @@ def generate_technique_content(technique_metadata):
   technique = add_default_values_technique_metadata(technique_metadata)
 
   content = []
-  for key in [ 'name', 'description', 'version' ]:
-    content.append('# @'+ key +" "+ technique[key])
+  for metadata_key in [ 'name', 'description', 'version' ]:
+    # Add commentary for each new line in the metadata
+    metadata_value = technique[metadata_key].replace("\n", "\n# ")
```
fanf: what about \r ? does it matter ? And other special chars ?

PR #209, opened 2015-06-24
handle weird cases for service_check_Started_at_boot

FlorianHeigl:
let service_check_started_at_boot also detect 'boot' only services on SLES, like i.e. boot.lvm.


peckpeck:
Good, thank you

PR #210, opened 2015-06-24
Service disabled boot

FlorianHeigl:
This allows to check that chkconfig servicename off has been set, needed to verify a service is in fact disabled.
This allows to disable a service at boot time.


amousset:
Done in #534

PR #211, opened 2015-06-25
Fixes #6796: generic methods permissions (non recursive) is indeed recursive

amousset:
https://www.rudder-project.org/redmine/issues/6796


ncharles:
\o/

PR #213, opened 2015-07-09
Fixes #6415: use, when available, proper tools for comparing versions…

ncharles:
… of package; and resynchronise cfengine stdlib

### DIFF
--- a/tree/20_cfe_basics/cfengine/bundles.cf
+++ b/tree/20_cfe_basics/cfengine/bundles.cf
@@ -1,5 +1,5 @@
 ############################################################################
-#  Copyright (C) Cfengine AS
+#  Copyright (C) CFEngine AS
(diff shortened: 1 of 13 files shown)


ncharles:
I tried to keep the modification to the minimum:
- compatible with CFEngine 3.5
- don't change behaviour
- minor modifications to ncf-lib (i decided not to use the path there, since the commit was already fairly big)

Kegeruneku:
Do we really need ncf_generic now that we are in sync again with the fixed stdlib ?

Kegeruneku:
Yes, it seems we do due to the removal of patch_command on Zypper handling (that made Zypper go mad and log tons of useless crap...).

Kegeruneku:
Accepted :+1:

PR #215, opened 2015-07-10
Fixes #6977: Add generic methods to allow convenient data management

amousset:
https://www.rudder-project.org/redmine/issues/6977

### DIFF
--- a/tree/30_generic_methods/data_merge.cf
+++ b/tree/30_generic_methods/data_merge.cf
@@ -1,0 +1,42 @@
+#####################################################################################
+# Copyright 2013 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name Merge data containers
+# @description Merges two data containers, the first is the default.
+#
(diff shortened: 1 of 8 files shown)


tree/30_generic_methods/variable_string.cf
```diff
@@ -0,0 +1,44 @@
...
+#
+# @parameter prefix Variable prefix
+# @parameter name   Name of the created variable
+# @parameter value  Value of the variable
+#
+# @class_prefix variable_string
+# @class_parameter name
+# This bundle will define a class variable_string_${prefix}_${name}_{kept,repaired,not_ok,ok,reached}
+
+bundle agent variable_string(prefix, name, value)
```
peckpeck: Couldn't we use the same value for prefix and for name ? Or what is the added value of prefix ?

tree/30_generic_methods/data_read.cf
```diff
@@ -0,0 +1,47 @@
...
+# @description Read a configuration file into a data variable
+#
+# @parameter name  Name of the created variable
+# @parameter file  Path of the configuration file
+#
+# @class_prefix file
+# @class_parameter file
+# This bundle will define a class data_read_${file}_{kept,repaired,not_ok,ok,reached}
+
+bundle agent data_read(name, file)
```
peckpeck: I'd keep this one

tree/30_generic_methods/data_merge.cf
```diff
@@ -0,0 +1,42 @@
...
+#
+# @parameter name  Name of the created variable
+# @parameter default_data Default data
+# @parameter user_data User data
+#
+# @class_prefix data_merge
+# @class_parameter name
+# This bundle will define a class data_merge_${name}_{kept,repaired,not_ok,ok,reached}
+
+bundle agent data_merge(name, default_data, user_data)
```
peckpeck: I'd replace this one with data_override with the same parameters as data_read 
But it would imply having mutable variables ...

tree/30_generic_methods/variable_string.cf
amousset: We could, it's just a "poor man's namespace"

PR #216, opened 2015-07-10
Fixes #6979: We should append .service to service name when calling systemctl

Kegeruneku:
https://www.rudder-project.org/redmine/issues/6979


peckpeck:
OK Good

PR #220, opened 2015-07-27
Fixes #7035: script tools/ncf-clone-test.sh use a bad target in makefile

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7035


peckpeck:
OK Good

PR #223, opened 2015-08-25
Fixes #5550: correct the classes confitiion when using quotes

ncharles:


VinceMacBuche:
It misses some tests :D but you already know that :)

ncharles:
Sorry, I forgot to include them

ncharles:
PR  updated !

PR #225, opened 2015-09-03
Fixes #7160: Canonify all class prefix

peckpeck:
https://www.rudder-project.org/redmine/issues/7160


ncharles:
does this really represent *all* the class prefix to canonize ?

peckpeck:
Indeed, I had just opened one file randomly when I found it and assumed it was more widespread

PR #226, opened 2015-09-04
Fixes #7156: Modify generic methods to define and use the new class_prefix

peckpeck:
https://www.rudder-project.org/redmine/issues/7156

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -355,6 +355,40 @@ body classes classes_generic_return_codes(x, kept_return_code, repaired_return_c
   promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
 }
 
+# WARNING !!!
+# For new class prefix migration only, this will be removed when migration is over
+# Define x and y prefixed/suffixed with promise outcome
+body classes classes_generic_two(x,y)
+{
+  !has_promiser_stack::
+    promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
+    repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
+    repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
+    repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
+    promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
+
+  has_promiser_stack::
+    promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached", 
+                          "promise_repaired_$(y)", "$(y)_repaired", "$(y)_ok", "$(y)_reached" };
+    repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached",
+                       "repair_failed_$(y)", "$(y)_failed", "$(y)_not_ok", "$(y)_error", "$(y)_not_kept", "$(y)_not_repaired", "$(y)_reached" };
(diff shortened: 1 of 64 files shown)


tree/30_generic_methods/file_copy_from_local_source.cf
```diff
@@ -28,6 +28,11 @@
...
 bundle agent file_copy_from_local_source(source, destination)
 {
+  vars:
+     "class_prefix"        string => canonify("file_copy_from_local_source_${destination}");
+     "new_class_prefix"    string => canonify(join("_", "this.callers_promisers"));
+
   methods:
-      "placeholder" usebundle => file_copy_from_local_source_recursion("${source}", "${destination}", "0");
+      "copy without recursion" usebundle => file_copy_from_local_source_recursion("${source}", "${destination}", "0");
+      "new result classes"     usebundle => _classes_copy("${class_prefix}", "${new_class_prefix}");
```
ncharles: i'm not sure of the result here.
If you call twice file_copy_from_local_source_recursion, with same destination, you won't be able to differenciate the results (the class_prefix will be the same)

peckpeck: The second time you call the file_copy_from_local_source method, you will have a different promiser, which will give a different stack from here, hence a different new_class_prefix.
We will still have a common class_prefix, but we we cannot get rid of it until we are sure that the promiser stack feature is supported.

ncharles: _class_copy cpy class_prefix (old value) into new_class_prefix (new value)
So the new promiser will be overriden by the old one (no way to distringuish if it is the previous call or this call that defined the _error or the _success)

peckpeck: Oh yes, then we'll have to define the classes at the source, ie define both in a new body classes

peckpeck:
PR updated

peckpeck:
Do not merge now, I will rename class_prefix to old_class_prefix and new_class_prefix to class_prefix

peckpeck:
PR updated

tree/30_generic_methods/file_copy_from_local_source.cf
```diff
@@ -28,6 +28,11 @@
...
 bundle agent file_copy_from_local_source(source, destination)
 {
+  vars:
+     "old_class_prefix" string => canonify("file_copy_from_local_source_${destination}");
+     "class_prefix"     string => canonify(join("_", "this.callers_promisers"));
+
   methods:
-      "placeholder" usebundle => file_copy_from_local_source_recursion("${source}", "${destination}", "0");
+      "copy without recursion" usebundle => file_copy_from_local_source_recursion("${source}", "${destination}", "0");
+      "new result classes"     usebundle => _classes_copy("${class_prefix}_copy_without_recursion", "${class_prefix}");
```
ncharles: shoudln't this call a logger ?

peckpeck: No, because the sub method already does it.
The logger refactoring will occur in a future PR.

peckpeck:
PR updated

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -355,6 +355,40 @@ body classes classes_generic_return_codes(x, kept_return_code, repaired_return_c
   promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
 }
 
+# WARNING !!!
+# For new class prefix migration only, this will be removed when migration is over
+# Define x and y prefixed/suffixed with promise outcome
+body classes classes_generic_two(x,y)
```
ncharles: x and y are really bad names :(

tree/30_generic_methods/file_copy_from_local_source.cf
ncharles: oh, ok !

tree/20_cfe_basics/ncf_lib.cf
peckpeck: This is merely a copy of the stdlib code.

tree/30_generic_methods/file_ensure_lines_present.cf
```diff
@@ -28,6 +28,11 @@
...
 bundle agent file_ensure_lines_present(file, lines)
 {
+  vars:
+      "old_class_prefix" string => canonify("file_ensure_lines_present_${file}");
+      "class_prefix"     string => canonify(join("_", "this.callers_promisers"));
+
   methods:
-      "wrapper" usebundle  => file_enforce_content("${file}", "${lines}", "false");
+      "enforce lines content" usebundle => file_enforce_content("${file}", "${lines}", "false");
+      "new result classes"    usebundle => _classes_copy("${class_prefix}_enforce_lines_content", "${class_prefix}");
```
ncharles: ${class_prefix}_enforce_lines_content will not be defined if you don't have the stack; what is the expected behaviour in this case?
old_class_prefix is not used, so should not be defined

peckpeck: If you don't have the stack class_prefix will be empty. _classes_copy detects this and doesn't do anything.

Indeed, old_class_prefix can be removed

PR #231, opened 2015-09-10
Fixes #7182: Deprecate methods that should never be used

peckpeck:
https://www.rudder-project.org/redmine/issues/7182


peckpeck:
PR updated

peckpeck:
PR replaced by https://github.com/Normation/ncf/pull/233

PR #232, opened 2015-09-11
Fixes #7186: ncf rpm package is missing asciidoc build dependencies

Kegeruneku:
https://www.rudder-project.org/redmine/issues/7186


peckpeck:
OK Good

PR #237, opened 2015-09-15
Fixes #7195: Canonify in ncf doesn't work like cfengine does

peckpeck:
https://www.rudder-project.org/redmine/issues/7195

### DIFF
--- a/tools/ncf_rudder.py
+++ b/tools/ncf_rudder.py
@@ -71,9 +71,17 @@ def canonify_expected_reports(expected_reports, dest):
 
     # Replace the second field with a canonified version of itself (a la CFEngine)
     fields = line.strip().split(";;")
-    regex = re.compile("[^a-zA-Z0-9_]", flags=re.UNICODE )
-    fields[1] = regex.sub("_", fields[1])
+    fields[1] = canonify(fields[1])
     dest_file.write(";;".join(fields) + "\n")
+
+def canonify(string):
+  # String should be unicode string (ie u'') which is the case if they are read from files opened with encoding="utf-8".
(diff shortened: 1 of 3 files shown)


tools/ncf_rudder.py
```diff
@@ -70,10 +70,17 @@ def canonify_expected_reports(expected_reports, dest):
...
-    regex = re.compile("[^a-zA-Z0-9_]", flags=re.UNICODE )
-    fields[1] = regex.sub("_", fields[1])
+    fields[1] = canonify(fields[1])
     dest_file.write(";;".join(fields) + "\n")
 
+def canonify(string):
+  # to match cfengine behaviour we need to treat utf8 as if it was ascii (see #7195)
+  # python2 uses iso strings here, but python3 uses utf8
+  if sys.version_info[0] != 2:
+    string = string.encode("utf-8").decode("iso-8859-1")
```
ncharles: isn't the iso-8859-1 dependant on the machine locale, or user input ?

ncharles: this looks really dodgy to me to convert to ise-8859-1

peckpeck: It is not dependent on the locale.
The "utf-8" is dependent on how ncf builder processes strings. And i'm pretty sure it is always utf8.
The "iso-8859-1" could be any 1 byte encoding, it is here to make sure python3 treats it as a string and not byte array so that we can apply a regex on it.

ncharles:
where are the unit tests ?

peckpeck:
PR updated

peckpeck:
PR updated

VinceMacBuche:
@peckpeck  can you rebase your PR please ? :)

peckpeck:
PR updated

PR #240, opened 2015-09-21
Fixes #7214: Do not consider 'deprecated' as a mandatory generic methods metadata

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7214


peckpeck:
OK Good

PR #241, opened 2015-09-21
Fixes #7215: Fix tests after 7182, 7158 and 7214

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7215


peckpeck:
OK Good

PR #243, opened 2015-09-22
Fixes #7217: syntax error in rudder-clone-test.sh

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7217


peckpeck:
OK Good

PR #248, opened 2015-10-03
fix for bug #7226

v92:
variable typo fix


peckpeck:
Thank you @v92 for your contribution, I would like to merge it.

But it need a proper comment to make our bugtracker happy ie start with "Fixes: #7226"
And it should aim the branch v0.x because we want to maintain a goof quality even for older versions.

So can you amend your commit with a new comment and make a new PR ?

Thanks!

PR #258, opened 2015-10-27
Add generic method to manage key = value in INI section

amousset:


amousset:
https://www.rudder-project.org/redmine/issues/6621

peckpeck:
The commit should have had a ref to the issue id.

PR #260, opened 2015-10-30
Fixes #7334: resynchronise with CFEngine stdlib

ncharles:


peckpeck:
OK, I trust you on that one!

PR #265, opened 2015-11-22
Fixes #7427: Invalid class parameter for command_execution_result.cf …

ncharles:
…on branch master


VinceMacBuche:
I did it in a merge lately ... why is it still open ? :(

peckpeck:
Trying a merge to see what the problem is

PR #266, opened 2015-11-23
Fixes #7434: Add edit-default with backup on all edit files

peckpeck:
https://www.rudder-project.org/redmine/issues/7434

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -375,5 +375,5 @@ body edit_defaults ncf_empty_select(select)
 {
         empty_file_before_editing => "${select}";
         max_file_size             => "1024000";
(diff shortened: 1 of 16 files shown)


tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -375,5 +375,5 @@ body edit_defaults ncf_empty_select(select)
 {
         empty_file_before_editing => "${select}";
         max_file_size             => "1024000";
-        edit_backup               => "false";
+        edit_backup               => "true";
```
ncharles: you should add that it is timestamped based

ncharles: edit_backup               => "timestamp";

peckpeck:
PR updated

PR #269, opened 2015-11-23
Fixes #7436: Techniques that call _logger, _logger_default or logger_rudder will break because these methods take 4 arguments now and no upgrade path has been planned

jooooooon:
https://www.rudder-project.org/redmine/issues/7436

### DIFF
--- a/tree/30_generic_methods/_log_default.cf
+++ b/tree/30_generic_methods/_log_default.cf
@@ -1,5 +1,5 @@
 #####################################################################################
-# Copyright 2013 Normation SAS
+# Copyright 2013-5 Normation SAS
 #####################################################################################
 #
 # This program is free software: you can redistribute it and/or modify
@@ -16,7 +16,7 @@
 #
 #####################################################################################
 
-# @name Logger default
+# @name Log default
 # @description Standard default logging output
 #
 # @parameter message              The common part of the message to display
@@ -29,14 +29,14 @@
 # The three states are kept, repaired and not_ok
 # (as defined in the classes_generic of the cfengine_stdlib)
 
-bundle agent _logger_default(message, old_class_prefix, origin_class_prefix, args)
+bundle agent _log_default(message, old_class_prefix, origin_class_prefix, args)
 {
   vars:
     "canonified_prefix" string => canonify("${old_class_prefix}");
 
     "promiser_count" int => length("this.callers_promisers");
 
-    # head(length-2) -> remove logger call stack (_logger + _logger_default)
+    # head(length-2) -> remove logger call stack (_log + _log_default)
     "nologger_len" string => eval("${promiser_count}-2", "math", "infix");
     "promisers_list_nologger" slist => sublist("this.callers_promisers", "head", "${nologger_len}");
 
--- a/tree/30_generic_methods/_logger.cf
+++ b/tree/30_generic_methods/_logger.cf
@@ -21,15 +21,13 @@
 #
 # @parameter message              The common part of the message to display
 # @parameter old_class_prefix     The prefix of the class for different states (0.x version)
-# @parameter origin_class_prefix  The prefix of the class for different states (1.x version)
-# @parameter args                 The arguments used to call the generic method (slist)
 #
 # @class_prefix
 # @class_parameter old_class_prefix
 # The three states are kept, repaired and not_ok
 # (as defined in the classes_generic of the cfengine_stdlib)
 
-bundle agent _logger(message, old_class_prefix, origin_class_prefix, args)
+bundle agent _logger(message, old_class_prefix)
 {
   vars:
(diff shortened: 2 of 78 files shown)


tree/30_generic_methods/_logger.cf
```diff
@@ -31,9 +31,13 @@
 
 bundle agent _logger(message, old_class_prefix, origin_class_prefix, args)
```
ncharles: shouldn't you change the signature of the method ?

tree/30_generic_methods/log_rudder.cf
```diff
@@ -0,0 +1,282 @@
...
+# There is one case it can conflict with the orignal one :
+#  - the expected report contains one line with a v2 style technique that matches current call
+#  - the expected report contains one line with an unrelated v1 style technique that matches the current call too
+# The answer is that you should re-run rudderify on all your techniques when you upgrade
+bundle agent _log_rudder_v2(expected_reports_source, message, class_prefix, args)
+{
+  vars:
+      # head(length-3) -> remove log call stack (_log + log_rudder + _log_rudder or _logger + _log + log_rudder + _log_rudder or logger_rudder + log_rudder + _log_rudder)
+      "promiser_count"  int => length("this.callers_promisers");
+      "nologger_len" string => eval("${promiser_count}-3", "math", "infix");
```
ncharles: this is 3.6 specific

ncharles: oh, this is only a file rename !

jooooooon:
PR updated

tree/30_generic_methods/log_rudder.cf
jooooooon: Yes, but we will require users to update away from 3.5 agents (Rudder 2.10) before upgrading.

tree/30_generic_methods/_logger.cf
jooooooon: Oops. Well spotted, thanks. Fixed.

tree/30_generic_methods/_log_default.cf
```diff
@@ -0,0 +1,74 @@
...
+
+# @name Log default
+# @description Standard default logging output
+#
+# @parameter message              The common part of the message to display
+# @parameter old_class_prefix     The prefix of the class for different states (0.x version)
+# @parameter origin_class_prefix  The prefix of the class for different states (1.x version)
+# @parameter args                 The arguments used to call the generic method (slist)
+#
+# @class_prefix _logger_default
```
ncharles: why the class prefix ? i'm pretty sure there's none of them

tree/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
```diff
@@ -32,7 +32,10 @@
 bundle agent file_ensure_key_value_present_in_ini_section(file, section, name, value)
 {
   vars:
-      "class_prefix" string => canonify("file_ensure_key_value_present_in_ini_section_${file}");
+      "old_class_prefix" string => canonify("file_ensure_key_value_present_in_ini_section_${file}");
+      "class_prefix"      string => canonify(join("_", "this.callers_promisers"));
+      "args"               slist => { "${file}", "${section}", "${name}", "${value}" };
```
ncharles: i don't want to be nitpicking, but could you align the => ?

tree/30_generic_methods/_log_default.cf
jooooooon: > why the class prefix ? i'm pretty sure there's none of them

All bundles must have a class_prefix. It's our internal style tests' policy. If we don't have one, tests will fail.

Also, there has always been one in this file, I just copied it without changing this line (see the individual commits, rather than the general overview)

tree/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
jooooooon: On 11/23/2015 08:23 PM, Nicolas Charles wrote:
>
> In
> tree/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf <https://github.com/Normation/ncf/pull/269#discussion_r45647722>:
>
> > @@ -32,7 +32,10 @@
> >  bundle agent file_ensure_key_value_present_in_ini_section(file, section, name, value)
> >  {
> >    vars:
> > -      "class_prefix" string => canonify("file_ensure_key_value_present_in_ini_section_${file}");
> > +      "old_class_prefix" string => canonify("file_ensure_key_value_present_in_ini_section_${file}");
> > +      "class_prefix"      string => canonify(join("_", "this.callers_promisers"));
> > +      "args"               slist => { "${file}", "${section}", "${name}", "${value}" };
>
> i don't want to be nitpicking, but could you align the => ?
>
>

Whoops. Done.

PR #272, opened 2015-11-24
Fixes #7461: classes_* methods should refuse to copy classes to a prefix which is an empty string

jooooooon:
https://www.rudder-project.org/redmine/issues/7461

### DIFF
--- a/tests/acceptance/30_generic_methods/classes_copy.cf
+++ b/tests/acceptance/30_generic_methods/classes_copy.cf
@@ -37,9 +37,10 @@ bundle agent test
     "ph2" usebundle => file_ensure_lines_absent("${init.file}", "sample one"); # should get success
     "ph3" usebundle => file_ensure_lines_absent("/dev/zero", "sample three"); # should get failed
 
-    "ph1"  usebundle => _classes_copy("file_ensure_lines_absent_${init.file_canon}","global_classA");
     "ph2"  usebundle => _classes_copy("file_ensure_lines_present_${init.file_canon}","global_classB");
+    "ph1"  usebundle => _classes_copy("file_ensure_lines_absent_${init.file_canon}","global_classA");
     "ph3"  usebundle => _classes_copy("file_ensure_lines_absent__dev_zero","global_classC");
+    "ph4"  usebundle => _classes_copy("file_ensure_lines_absent_${init.file_canon}","");
 
 }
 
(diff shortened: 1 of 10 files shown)


tests/acceptance/30_generic_methods/classes_copy.cf
```diff
@@ -37,9 +37,10 @@ bundle agent test
     "ph2" usebundle => file_ensure_lines_absent("${init.file}", "sample one"); # should get success
     "ph3" usebundle => file_ensure_lines_absent("/dev/zero", "sample three"); # should get failed
 
-    "ph1"  usebundle => _classes_copy("file_ensure_lines_absent_${init.file_canon}","global_classA");
     "ph2"  usebundle => _classes_copy("file_ensure_lines_present_${init.file_canon}","global_classB");
```
peckpeck: Why do you change the ordering ?

tree/30_generic_methods/_classes_copy.cf
```diff
@@ -33,6 +33,7 @@ bundle agent _classes_copy(source_prefix, destination_prefix)
       "prefix" slist => { "promise_kept", "promise_repaired", "repair_failed", "repair_denied", "repair_timeout" };
       "suffix" slist => { "repaired", "ok", "reached", "failed", "not_ok", "not_kept", "not_repaired", "denied", "timeout", "kept", "error" };
 
+    destination_ok::
```
peckpeck: In master, this is named destination_exists.
The merge will be easier if you use the same name

tests/acceptance/30_generic_methods/classes_copy.cf
jooooooon: Because the original ordering made no sense. Now, the order matches the order of the calls above.

jooooooon:
PR updated

peckpeck:
Good, merging

tree/30_generic_methods/_classes_success.cf
```diff
@@ -38,6 +39,10 @@ bundle agent _classes_success(class_prefix)
       "local_${suffix}"
         string     => "${class_prefix}_${suffix}",
         classes    => always("${class_prefix}_${suffix}");
+
+  classes:
+      "destination_exists" not => strcmp("${class_prefix}", "");
```
ncharles: shouldn't it report a warning if an empty string is passed?

peckpeck: In master we use this property in every generic method so that calls don't fail when this.caller_promisers doesn't exist. So it could be a warning, but not in the general case.

PR #273, opened 2015-11-24
Fixes #7463: classes_* methods should refuse to copy classes to a prefix which is an empty string (master branch)

jooooooon:
https://www.rudder-project.org/redmine/issues/7463


peckpeck:
Good, merging

PR #274, opened 2015-11-26
Fixes #7483: list_compatible_input doesn't list 3.6 specific files

peckpeck:
https://www.rudder-project.org/redmine/issues/7483

### DIFF
--- a/tree/10_ncf_internals/list-compatible-inputs
+++ b/tree/10_ncf_internals/list-compatible-inputs
@@ -21,8 +21,8 @@ export PATH
 
 # split version numbers
 version_regex='\([0-9][0-9]*\)\.\([0-9][0-9]*\).*'
-cfengine_major=`printf ${cfengine_version} | sed -e "s/${version_regex}/\\1/"`
-cfengine_minor=`printf ${cfengine_version} | sed -e "s/${version_regex}/\\2/"`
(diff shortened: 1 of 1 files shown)


tree/10_ncf_internals/list-compatible-inputs
```diff
@@ -21,8 +21,8 @@ export PATH
 
 # split version numbers
 version_regex='\([0-9][0-9]*\)\.\([0-9][0-9]*\).*'
-cfengine_major=`printf ${cfengine_version} | sed -e "s/${version_regex}/\\1/"`
-cfengine_minor=`printf ${cfengine_version} | sed -e "s/${version_regex}/\\2/"`
+cfengine_major=`printf "${cfengine_version}\n" | sed -e "s/${version_regex}/\\1/"`
+cfengine_minor=`printf "${cfengine_version}\n" | sed -e "s/${version_regex}/\\2/"`
```
jooooooon: Wouldn't it be simpler just to use ```echo``` ?

peckpeck: echo is not posix and doesn't work on some os

jooooooon: Ah, fair enough :)

PR #275, opened 2015-11-28
Fixes #7500: Errors in inform output about 'assumed integer value' due to time_lib.cf

jooooooon:
https://www.rudder-project.org/redmine/issues/7500


peckpeck:
replaced bu #282

PR #276, opened 2015-12-01
Fixes #7521: Tests do not work on AIX

jooooooon:
https://www.rudder-project.org/redmine/issues/7521

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: clean test doc
 
 test:
-	type fakeroot 2>/dev/null || { echo "fakeroot is required but not found." ; exit 1 ; }
+	[ `id | cut -d\( -f2 | cut -d\) -f1` = 'root' ] || type fakeroot 2>/dev/null || { echo "Not running as root and fakeroot not found." ; exit 1 ; }
 	cd tests/style/ && ./testall
 	cd tests/unit/ && ./testall
 	cd tests/acceptance/ && ./testall --no-network
@@ -44,7 +44,7 @@ clean:
 	rm -rf tests/acceptance/workdir/
 	rm -f doc/all_generic_methods.txt
 	rm -f doc/generic_methods.md
-	find $(CURDIR) -iname "*.pyc" -delete
+	find $(CURDIR) -name "*.[pP][yY][cC]" -exec rm "{}" \;
 
 distclean: clean
 
--- a/tests/acceptance/30_generic_methods/file_check_exists.cf
+++ b/tests/acceptance/30_generic_methods/file_check_exists.cf
@@ -33,7 +33,7 @@ bundle agent init
   files:
     "${file}"
       create => "true",
-      perms  => mog("000", "root", "root");
+      perms  => mog("000", "root", "0");
 }
 
 #######################################################
--- a/tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
+++ b/tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
@@ -30,7 +30,7 @@ bundle agent init
 
     "mode"                   string => "644";
     "owner"                  string => "root";
-    "group"                  string => "root";
+    "group"                  string => "0";
 
   files:
     "${source_file}"
@@ -52,9 +52,10 @@ bundle agent test
 bundle agent check
 {
   vars:
-    "permissions_test_mode"  string => "/usr/bin/test $(/usr/bin/stat -c %a ${init.destination_file}) = \"${init.mode}\"";
-    "permissions_test_owner" string => "/usr/bin/test $(/usr/bin/stat -c %U ${init.destination_file}) = \"${init.owner}\"";
-    "permissions_test_group" string => "/usr/bin/test $(/usr/bin/stat -c %G ${init.destination_file}) = \"${init.group}\"";
+    "owner_id"                  int => getuid("${init.owner}");
+    "permissions_test_mode"  string => "/usr/bin/test $(${test_utils.file_perms} ${init.destination_file}) = \"${init.mode}\"";
+    "permissions_test_owner" string => "/usr/bin/test $(${test_utils.file_owner} ${init.destination_file}) = \"${owner_id}\"";
+    "permissions_test_group" string => "/usr/bin/test $(${test_utils.file_group} ${init.destination_file}) = \"${init.group}\"";
 
   classes:
     # By default, file_copy_from_local_source_type_recursion should create the file if it doesn't exist
--- a/tests/acceptance/30_generic_methods/file_ensure_lines_present_test_already_present.cf
+++ b/tests/acceptance/30_generic_methods/file_ensure_lines_present_test_already_present.cf
@@ -26,7 +26,7 @@ bundle agent init
     "tmp"         string => getenv("TEMP", 1024);
     "file"        string => "${tmp}/test";
     "file_canon"  string => canonify("${file}");
-    "line_to_add" string => "This is a test line!";
+    "line_to_add" string => "This is a test line.";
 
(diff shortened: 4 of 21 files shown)


Makefile
```diff
@@ -44,7 +44,7 @@ clean:
 	rm -rf tests/acceptance/workdir/
 	rm -f doc/all_generic_methods.txt
 	rm -f doc/generic_methods.md
-	find $(CURDIR) -iname "*.pyc" -delete
+	find $(CURDIR) -name "*.[pP][yY][cC]" -exec rm "{}" \;
```
jooooooon: ```-iname``` does not exist on AIX 5.3's find

jooooooon:
PR updated

tests/acceptance/30_generic_methods/file_check_exists.cf
```diff
@@ -33,7 +33,7 @@ bundle agent init
   files:
     "${file}"
       create => "true",
-      perms  => mog("000", "root", "root");
+      perms  => mog("000", "root", "0");
```
jooooooon: "root" group does not exist on AIX, so we use "0" instead. A style test is added below so this is automatically detected.

tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
```diff
@@ -52,9 +52,10 @@ bundle agent test
 bundle agent check
 {
   vars:
-    "permissions_test_mode"  string => "/usr/bin/test $(/usr/bin/stat -c %a ${init.destination_file}) = \"${init.mode}\"";
-    "permissions_test_owner" string => "/usr/bin/test $(/usr/bin/stat -c %U ${init.destination_file}) = \"${init.owner}\"";
-    "permissions_test_group" string => "/usr/bin/test $(/usr/bin/stat -c %G ${init.destination_file}) = \"${init.group}\"";
+    "owner_id"                  int => getuid("${init.owner}");
```
jooooooon: This is necessary because the new ```test_utils.file_owner``` command below returns a UID, not a user name.

tests/acceptance/30_generic_methods/file_ensure_lines_present_test_already_present.cf
```diff
@@ -26,7 +26,7 @@ bundle agent init
     "tmp"         string => getenv("TEMP", 1024);
     "file"        string => "${tmp}/test";
     "file_canon"  string => canonify("${file}");
-    "line_to_add" string => "This is a test line!";
+    "line_to_add" string => "This is a test line.";
```
jooooooon: Using "!" caused havoc because of the shell interpreting it in the test results.

tests/acceptance/30_generic_methods/file_ensure_lines_present_test_already_present.cf
```diff
@@ -48,16 +48,14 @@ bundle agent test
...
-    "nb_lines" 
-      string     => execresult("${line_exists_test}", "useshell"),
-      ifvarclass => canonify("file_ensure_lines_present_${init.file}_reached");
+    "line_exists_test" string => "/usr/bin/test `/bin/grep \"^${init.line_to_add}$\" \"${init.file}\" | wc -l` = '1'";
 
   classes:
     # By default, file_ensure_lines_present should create the file if it doesn't exist
     "file_exists" expression => fileexists("${init.file}");
     
-    "line_exists" expression => strcmp("${nb_lines}", "1");
```
jooooooon: The strings were not equal on AIX because ```wc -l``` returns a lot of leading spaces. Doing the comparison in shell means these spaces are ignored.

tests/acceptance/default_ncf.cf.sub
```diff
@@ -9,9 +9,26 @@ bundle common ncf_inputs
      # long string to be built, causing CFEngine to error out, so we rebuild
      # the full paths using maplist below
      "generic_ncf_files"
-        string => execresult("/bin/bash -c \"cd ${ncf_tree} && /usr/bin/find 10_ncf_internals/configuration.cf 10_ncf_internals/expand_configuration.cf 20_cfe_basics 30_generic_methods 40_it_ops_knowledge -name '*.cf' -printf '%p\n'\"", "noshell");
+        string => execresult("/bin/bash -c \"cd ${ncf_tree} && /usr/bin/find 10_ncf_internals/configuration.cf 10_ncf_internals/expand_configuration.cf 20_cfe_basics 30_generic_methods 40_it_ops_knowledge -name '*.cf' -print\"", "noshell");
```
jooooooon: ```find```'s ```-printf``` option is not available on AIX 5.3. However, it is not needed, ```-print``` is just fine, and what we already use in ```list-compatible-inputs```.

tests/acceptance/testall
```diff
@@ -66,12 +68,14 @@ LOG=test.log
 SUMMARY=summary.log
 XML=test.xml
 XMLTMP=xml.tmp
-BASE_WORKDIR="$(pwd)/workdir"
+BASE_WORKDIR="`pwd`/workdir"
```
jooooooon: Most changes in this file are just replace ```$(command)``` with ``` `command` ```

tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
peckpeck: Why do you sometime use getuid and on other tests, use owner=0 ?

tests/acceptance/30_generic_methods/file_from_template.cf
```diff
@@ -47,14 +47,14 @@ bundle agent test
 bundle agent check
 {
   vars:
-    "file_length_check" string => "/bin/cat ${init.destination_file} | wc -l";
+    "file_length_check" string => "/usr/bin/test `/bin/cat ${init.destination_file} | wc -l` = \"1\"";
```
peckpeck: double quotes are not strictly necessary

tests/acceptance/testall
```diff
@@ -504,21 +509,23 @@ START_TIME=$(unix_seconds)
   echo -n "" > "$XMLTMP"
 
 for test in $TESTS; do
-  echo "    "\<testcase name=\"$(basename $test)\" >> "$XMLTMP"
+  echo "    "\<testcase name=\"`basename $test`\" >> "$XMLTMP"
   echo "              "classname=\"$test\" >> "$XMLTMP"
+  set +e # don't fail on errors when running the tests - this is expected
```
peckpeck: Why not adding a || true on commands that can fail.
Then we still can disable set -e globally

tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
jooooooon: Because getuid fails if you pass it a number :-( 

I've tried to always use a bane for the user, but for the group I use 0 because the "root" group name is different across OSes

tests/acceptance/testall
jooooooon: Good idea!

tests/acceptance/30_generic_methods/file_copy_from_local_source_recursion.cf
peckpeck: You could directly change init.owner to be 0 like you did on other tests.

jooooooon: Ah actually I meant to do the opposite: never use 0 for the owner, because it's not necessary, and "root" is more well known. I'll homogenise these.

jooooooon:
PR updated

peckpeck:
OK, merging

PR #277, opened 2015-12-02
Fixes #7529: Add support for the SRC system on AIX for service management

jooooooon:
https://www.rudder-project.org/redmine/issues/7529

### DIFF
--- a/tests/acceptance/30_generic_methods/service_check_started_at_boot.cf
+++ b/tests/acceptance/30_generic_methods/service_check_started_at_boot.cf
@@ -1,6 +1,6 @@
 #######################################################
 #
-# Test if cron is started
+# Test if cron is started at boot
 #
 #######################################################
 
@@ -24,9 +24,7 @@ bundle agent init
   vars:
     redhat::
       "service_name"  string => "crond";
-    aix::
-      "service_name"  string => "sshd";
-    !(redhat|aix)::
+    !redhat::
       "service_name"  string => "cron";
 }
 
@@ -35,7 +33,7 @@ bundle agent init
 bundle agent test
 {
   methods:
-    "ph1" usebundle => service_check_running("${init.service_name}");
+    "ph1" usebundle => service_check_started_at_boot("${init.service_name}");
 }
 
 #######################################################
@@ -43,9 +41,9 @@ bundle agent test
 bundle agent check
 {
   classes:
-    "ok_success"  expression => "(promise_kept_service_check_running_${init.service_name}.service_check_running_${init.service_name}_kept.service_check_running_${init.service_name}_ok.service_check_running_${init.service_name}_not_repaired.service_check_running_${init.service_name}_reached)";
-    "ok_repaired" expression => "!(promise_repaired_service_check_running_${init.service_name}|service_check_running_${init.service_name}_repaired)";
-    "ok_error"    expression => "!(service_check_running_${init.service_name}_not_kept|repair_failed_service_check_running_${init.service_name}|service_check_running_${init.service_name}_failed|service_check_running_${init.service_name}_not_ok)";
+    "ok_success"  expression => "(promise_kept_service_check_started_at_boot_${init.service_name}.service_check_started_at_boot_${init.service_name}_kept.service_check_started_at_boot_${init.service_name}_ok.service_check_started_at_boot_${init.service_name}_not_repaired.service_check_started_at_boot_${init.service_name}_reached)";
+    "ok_repaired" expression => "!(promise_repaired_service_check_started_at_boot_${init.service_name}|service_check_started_at_boot_${init.service_name}_repaired)";
+    "ok_error"    expression => "!(service_check_started_at_boot_${init.service_name}_not_kept|repair_failed_service_check_started_at_boot_${init.service_name}|service_check_started_at_boot_${init.service_name}_failed|service_check_started_at_boot_${init.service_name}_not_ok)";
     "ok"          and        => { "ok_success", "ok_repaired", "ok_error" };
 
   reports:
--- a/tests/acceptance/30_generic_methods/unsafe/service_management.cf
+++ b/tests/acceptance/30_generic_methods/unsafe/service_management.cf
@@ -25,7 +25,9 @@ bundle agent init
   vars:
     redhat::
       "service" string => "sshd";
-    !redhat::
+    aix::
+      "service" string => "syslogd";
+    !(redhat|aix)::
       "service" string => "ssh";
 }
 
@@ -36,7 +38,7 @@ bundle agent test
 
   methods:
 
-      "ph" usebundle => service_management("${init.service}", "reload");
+      "ph" usebundle => service_action("${init.service}", "reload");
 
 }
 
@@ -45,7 +47,7 @@ bundle agent test
 bundle agent check
 {
   classes:
-      "ok" expression => "service_management_${init.service}_reached.service_management_${init.service}_repaired.!service_management_${init.service}_kept.!service_management_${init.service}_error";
+      "ok" expression => "service_action_${init.service}_reached.service_action_${init.service}_repaired.!service_action_${init.service}_kept.!service_action_${init.service}_error";
 
   reports:
     ok::
--- a/tree/30_generic_methods/service_action.cf
+++ b/tree/30_generic_methods/service_action.cf
@@ -39,7 +39,15 @@ bundle agent service_action(service_name, action)
 
       "action_command"            string => "${paths.path[service]} ${service_name} ${action}";
 
-    !systemctl_utility_present.!service_utility_present.init_d_directory_present::
+    !systemctl_utility_present.!service_utility_present.startsrc_utility_present::
+      "svc_action_cmd[restart]"   string => "/usr/bin/stopsrc -s ${service_name} && until /usr/bin/lssrc -s ${service_name} | ${paths.grep} -q inoperative; do ${paths.perl} -e 'select(undef,undef,undef,.25)'; done; /usr/bin/startsrc -s ${service_name}";
+      "svc_action_cmd[refresh]"   string => "/usr/bin/refresh -s ${service_name}";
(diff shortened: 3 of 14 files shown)


tests/acceptance/30_generic_methods/unsafe/service_management.cf
```diff
@@ -36,7 +36,7 @@ bundle agent test
 
   methods:
 
-      "ph" usebundle => service_management("${init.service}", "reload");
+      "ph" usebundle => service_action("${init.service}", "reload");
```
jooooooon: This bundle was renamed a while ago. It doesn't exist.

tree/30_generic_methods/service_ensure_running_path.cf
```diff
@@ -36,22 +36,30 @@ bundle agent service_ensure_running_path(service_name, service_path)
 
     "class_prefix"                 string => "service_ensure_running_${canonified_service_name}";
 
+  classes:
+    # This method can be called as a substitution if service_ensure_running is called directly, with service_path=service_name
+    # In this case, we don't need to call service_check_running_ps but can use better OS-specific methods if available
+    "real_path"                       not => strcmp("${service_name}", "${service_path}");
```
jooooooon: This is rather ugly. The problem is that ```service_ensure_running(x)``` calls ```service_ensure_running_path(x, x)```, so there is no other implementation than this one. That was not a problem until now, because with only one implementation, there were no choices to be made. But now, there is a "real" ```service_check_running``` method that uses src, not ps. So we need to make a choice... This is kinda like backpedalling, though.

tree/30_generic_methods/service_action.cf
```diff
@@ -39,7 +39,14 @@ bundle agent service_action(service_name, action)
 
       "action_command"            string => "${paths.path[service]} ${service_name} ${action}";
 
-    !systemctl_utility_present.!service_utility_present.init_d_directory_present::
+    !systemctl_utility_present.!service_utility_present.startsrc_utility_present::
+      "svc_action_cmd[restart]"   string => "/usr/bin/stopsrc -s ${service_name}; /usr/bin/startsrc -s ${service_name}";
```
ncharles: This method may not always work: stopsrc says that the service was requested to stop, not that it stoped
cf https://www.ibm.com/developerworks/community/blogs/brian/entry/scripting_the_stop_and_restart_of_src_controlled_processes_on_aix6?lang=en

the common workaround is:
stopsrc -s sshd
sleep 10
startsrc -s sshd

the tricky one (and hard to make all purpose) is stopsrc -s sshd && until lssrc -s sshd | grep -q inoperative; do echo "Waiting for sshd to stop"; perl -e 'select(undef,undef,undef,.25)'; done && startsrc -s sshd

jooooooon:
Please wait before merging - this is missing support for service_check_started_at_boot

jooooooon:
PR updated

jooooooon:
PR updated

jooooooon:
PR updated

tree/30_generic_methods/service_action.cf
jooooooon: OK, sounds good to me. I don't see why this is harder, though. Implementing.

jooooooon:
PR updated

jooooooon:
OK, this is ready for review and merge again!

tree/30_generic_methods/service_check_started_at_boot.cf
```diff
@@ -35,7 +35,10 @@ bundle agent service_check_started_at_boot(service_name)
     pass2.!systemctl_utility_present.chkconfig_utility_present::
       "command_to_check"   string => "${paths.path[chkconfig]} --list ${service_name} | grep -q -e 3:on -e B:on";
 
-    pass2.(!(systemctl_utility_present|chkconfig_utility_present)|broken_systemctl)::
+    pass2.!systemctl_utility_present.!chkconfig_utility_present.lsitab_utility_present::
+      "command_to_check"   string => "lsitab -a | egrep '^${service_name}:[0-9]+:(respawn|boot|bootwait|wait|once|initdefault|sysinit):'";
```
ncharles: could you use the full paths to the binaries ?

jooooooon: Wow, I'm amazed that CFEngine doesn't complain about this anymore!! Yes, of course, fixing

tests/acceptance/30_generic_methods/service_check_started_at_boot.cf
```diff
@@ -0,0 +1,51 @@
...
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "service_name"  string => "cron";
```
ncharles: is it always cron on all platform ? On my CentOS, i see crond in /etc/init.d

jooooooon:
PR updated

tests/acceptance/30_generic_methods/service_check_started_at_boot.cf
jooooooon: Good catch. Fixing.

jooooooon:
PR updated

tests/acceptance/30_generic_methods/service_check_running.cf
```diff
@@ -22,7 +22,12 @@ body common control
 bundle agent init
 {
   vars:
-    "service_name"  string => "cron";
+    redhat::
+      "service_name"  string => "crond";
+    aix::
+      "service_name"  string => "sshd";
```
peckpeck: Given that most of the time the machine doing the test will be controlled by someone like us you can use a use a less critical service like syslogd, snmpd or inetd.

tree/30_generic_methods/service_action.cf
```diff
@@ -39,7 +39,15 @@ bundle agent service_action(service_name, action)
 
       "action_command"            string => "${paths.path[service]} ${service_name} ${action}";
 
-    !systemctl_utility_present.!service_utility_present.init_d_directory_present::
+    !systemctl_utility_present.!service_utility_present.startsrc_utility_present::
+      "svc_action_cmd[restart]"   string => "/usr/bin/stopsrc -s ${service_name} && until lssrc -s ${service_name} | grep -q inoperative; do perl -e 'select(undef,undef,undef,.25)'; done; /usr/bin/startsrc -s ${service_name}";
```
peckpeck: Why do you use full path for startsrc and not for the other commands ?

tree/30_generic_methods/service_check_running_src.cf
```diff
@@ -0,0 +1,56 @@
...
+# @name Service check running with SRC
+# @description Check if a service is running using SRC (AIX System Resource Controller)
+#
+# @parameter service_name Subsystem name in SRC
+#
+# @class_prefix service_check_running
+# @class_parameter service_name
+# This bundle will define a class service_check_running_${service_name}_{kept,not_ok}
+
+bundle agent service_check_running_src(service_name)
```
peckpeck: This is an AIX specific generic method, can't it be integrated directly into service_check_running with a class condition ?

tests/acceptance/30_generic_methods/service_check_running.cf
jooooooon: True, but this is just a check. No harm can be done.

tree/30_generic_methods/service_action.cf
jooooooon: This is a mistake. We have to use full paths, because CFEngine (at least <= 3.6) checks if the first "string" in a command in a valid binary. The rest of the line doesn't stricly *have* to use full paths but it is good practice to avoid errors when refactoring.

tree/30_generic_methods/service_check_running_src.cf
jooooooon: Well, it could, but my vision of ncf is that we should be working the opposite way. One of our core principles is KISS - keep it simple. Merging this code into the upper bundle means that bundle will quickly become unwieldy and complex. What usually happens first in this kind of "mega-bundle" is that class conditions become very complex, and then we start making mistakes.

For an example, take a look at the ```service_action``` method, where we have merged all possible implementations into one method, in particular this class condition:

```
systemctl_utility_present|service_utility_present|init_d_directory_present|startsrc_utility_present|(windows.is_valid_action)::
```

Or this one in ```service_check_started_at_boot```:

```
pass2.(!(systemctl_utility_present|chkconfig_utility_present|lsitab_utility_present)|broken_systemctl)::
```

I want to avoid this as far as possible.

I see a much simple approach where we have a generic method for each type of implementation (src, systemd, upstart, init.d, ...), and ```service_check_running ``` is simply a "router" to choose which implementation to use according to the OS we're on and the utilities available.

jooooooon:
PR updated

tree/30_generic_methods/service_check_running_src.cf
peckpeck: I think implementation detail and OS specific code should not be visible to the user of ncf API.
If we want to split bundles to make the implementation cleaner, this should go into 20_ and not into a generic method.

jooooooon: They should most definitely not go into 20_, because that is for body libraries.

The use of "_" would be ideal here, like _service_check_running_src. This hides methods from users.

However, we already a couple like this in ncf today:

* service_check_running_ps.cf
* service_ensure_running_path.cf

peckpeck: service with path and without have different parameters that cannot be guessed by looking at them and they are badly named since the path versions check processes not services.

jooooooon:
PR updated

tree/30_generic_methods/service_check_running_src.cf
jooooooon: OK, let's rename it _service_check_running_src then.

peckpeck:
Good, merging

PR #285, opened 2015-12-07
Fixes #7519: http_request_* can't fallback on wget and doesn't warn if curl is missing

jooooooon:
https://www.rudder-project.org/redmine/issues/7519


peckpeck:
OK merging

PR #286, opened 2015-12-07
Fixes #7553: ncf 1.x show debug messages with unknown promisers and args lists

jooooooon:
https://www.rudder-project.org/redmine/issues/7553


peckpeck:
OK, merging

PR #290, opened 2015-12-13
Fixes #7604: group_present and group_absent use non-existant class \"unix\"

jooooooon:
https://www.rudder-project.org/redmine/issues/7604

### DIFF
--- a/tests/acceptance/30_generic_methods/unsafe/group_present.cf
+++ b/tests/acceptance/30_generic_methods/unsafe/group_present.cf
@@ -61,12 +61,16 @@ bundle agent check
       "$(this.promise_filename) Pass";
     !ok::
       "$(this.promise_filename) FAIL";
-#    group1_ok::
-#      "group1_ok";
-#    group2_ok::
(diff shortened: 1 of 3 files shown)


tests/acceptance/30_generic_methods/unsafe/group_present.cf
```diff
@@ -61,12 +61,16 @@ bundle agent check
...
+    group2_ok::
+      "group2_ok";
+    ph1_ok::
+      "ph1_ok";
+    ph2_ok::
+      "ph2_ok";
+    cfengine::
+      "group_present_${init.group1}_kept" ifvarclass => "group_present_${init.group1}_kept";
+      "group_present_${init.group1}_repaired" ifvarclass => "group_present_${init.group1}_repaired";
+      "group_present_${init.group1}_error" ifvarclass => "group_present_${init.group1}_error";
```
jooooooon: This is just adding some debug to the test to make it easier to find out where it's failing, harmless

PR #292, opened 2015-12-16
Fixes #7626: Tehcnique editor center part has no overflow in an iframe

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7626


ncharles:
Thank you so much !

PR #295, opened 2015-12-21
rename class_prefix

frbayart:
rename class_prefix to package_verify_version to avoid conflit with package_install

### DIFF
--- a/tree/30_generic_methods/package_verify_version.cf
+++ b/tree/30_generic_methods/package_verify_version.cf
@@ -22,9 +22,9 @@
 # @parameter package_name    Name of the package to verify
 # @parameter package_version Version of the package to verify (can be "latest" for latest version)
 #
-# @class_prefix package_install
+# @class_prefix package_verify_version
 # @class_parameter package_name
-# This bundle will define a class package_install_${package_name}_{kept,repaired,not_ok,ok,reached}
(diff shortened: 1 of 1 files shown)


peckpeck:
This method calls package_install_version_cmp which effectively uses the package_install class prefix.
Removing the class prefix in the method metadata is not sufficient to change the prefix.

frbayart:
Ok so I'm lucky because my update fix my problem, I'm now able to use class condition on package verify.

What is the clean solution to avoid conflict between classname in packages methods ?

peckpeck:
There is no clean solution ready yet.
The problem is that they all use the same method in fine and this methods has a static class_prefix.
The solution is to have a class_prefix that depends on the call stack, that's why there is now and old_class_prefix and a new class_prefix in the code.

But this feature will only be available when the caller promiser patch will be merged into cfengine. 
We didn't include this patch in Rudder because we are not yet sure if it will be taken in the current form. The original PR is here https://github.com/cfengine/core/pull/2301

frbayart:
so we can only use one "package" methods by techniques in ncf editor.

This news is very disappointing !

peckpeck:
Sorry, this is a very longstanding bug in ncf that depends on the way it was designed and on a missing feature in cfengine.

PR #297, opened 2015-12-22
Fixes #7674: Adapt support for the SMF system on Solaris for service management to v1.0

peckpeck:
https://www.rudder-project.org/redmine/issues/7674


ncharles:
looks ok

PR #300, opened 2015-12-22
Fixes #7683: Syntax error in _service_check_running_smf.cf

ncharles:
https://www.rudder-project.org/redmine/issues/7683


peckpeck:
OK, merging

PR #301, opened 2015-12-28
Fixes #7693: service_check_running result classes broken on master branch

jooooooon:
https://www.rudder-project.org/redmine/issues/7693


peckpeck:
OK, master branch only

PR #302, opened 2015-12-28
Fixes #7694: service_check_running report misleading always says it uses 'ps'

jooooooon:
https://www.rudder-project.org/redmine/issues/7694


peckpeck:
OK Good

PR #303, opened 2015-12-28
Fixes #7695: service_ensure_running_path creates the wrong result classes on master

jooooooon:
https://www.rudder-project.org/redmine/issues/7695


tree/30_generic_methods/service_ensure_running_path.cf
```diff
@@ -74,7 +74,7 @@ bundle agent service_ensure_running_path(service_name, service_path)
         ifvarclass => "(real_path.!service_check_running_${canonified_service_path}_kept)|(!real_path.!service_check_running_${canonified_service_name}_kept)";
 
       "class if running"
-        usebundle  => _classes_success("${class_prefix}"),
+        usebundle  => _classes_success("${old_class_prefix}"),
```
peckpeck: Shouldn't we do both here ?

jooooooon: The new one is already done above, see line 63

peckpeck:
Good, merging

PR #304, opened 2016-01-04
Fixes #7724: Allow to drag and drop method call (move, create)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7724


fanf:
Some notes:
- there is an ugly box shadows around methods title (perhaps only when there is a missing mandatory parameter)
- the drag'n'drop does not work for the first item (the landing zone is missing)
- I'm not sure that there is still use for the moving arrows, and they don't invite to drag'n'drop and discover the new (much much much better) UX
- it misses something to make the user know he can drag'n'drop methods around. Perhaps a text or an hover notice or something
- on new techniques, the methods list should be already open, [add method] is a mandatory click without added value

Appart from that, IT'S JUST SO GOOD

VinceMacBuche:
PR updated

builder/js/ncf.js
```diff
@@ -103,7 +101,24 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location,
     }
   };
 
+  // Callback when an element is dropped on the list of method calls
+  // return the element that will be added, if false do not add anything
+  // We need to handle removal of previous element when moving an element
```
fanf: I think the comment last line is no more accurate

fanf:
It's much better, really nice now. 
The only things that I'm not sure about is the [add metchods] button that is greyed when a method is being edited, but that you can click to get back to the list of methods. I'm not sure I want something else, it's just strange to have a greyed button working. Perhaps letting its normal aspect would be better.

VinceMacBuche:
PR updated

PR #306, opened 2016-01-06
Fixes #7715: ncf editor and display problem

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7715


jooooooon:
This is important. Ping @fanf ?

PR #307, opened 2016-01-11
Fixes #7520: package_install* methods do not work outside of debian|redhat|SuSE Linux distributions and don't warn about it

jooooooon:
https://www.rudder-project.org/redmine/issues/7520

### DIFF
--- a/tree/30_generic_methods/package_install_version_cmp_update.cf
+++ b/tree/30_generic_methods/package_install_version_cmp_update.cf
@@ -17,7 +17,7 @@
 #####################################################################################
 
 # @name Package install version compare (update optional)
-# @description Install a package or verify if it is installed in a specific version, or higher or lower version than a version specified, optionally test update or not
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/package_install_version_cmp_update.cf
```diff
@@ -171,4 +171,9 @@ bundle agent package_install_version_cmp_update(package_name, version_comparator
       "Reports" usebundle => _logger("Verify that package ${package_name} is installed and up-to-date", "${class_prefix}");
     !immediate.version_not_specified.!update::
       "Reports" usebundle => _logger("Verify that package ${package_name} is installed", "${class_prefix}");
+
+    !debian.!redhat.!SuSE::
```
peckpeck: shouldn't we add (debian|redhat|suse) in previous conditions ?

jooooooon:
PR updated

tree/30_generic_methods/package_install_version_cmp_update.cf
jooooooon: Yes, absolutely! PR updated.

peckpeck:
Good, merging

PR #313, opened 2016-01-27
Fixes #7848: Using configuration value before defining it

amousset:
https://www.rudder-project.org/redmine/issues/7848


peckpeck:
I'd like the opinion of @ncharles on this

peckpeck:
OK, merging

PR #315, opened 2016-02-01
Fixes #7856: Use a mustache template from a string

peckpeck:
https://www.rudder-project.org/redmine/issues/7856


tree/30_generic_methods/file_from_string_mustache.cf
```diff
@@ -0,0 +1,46 @@
+#####################################################################################
+# Copyright 2014 Normation SAS
```
jooooooon: 2016 :)

tree/30_generic_methods/file_from_string_mustache.cf
```diff
@@ -0,0 +1,46 @@
...
+# @parameter destination     Destination file
+#
+# @class_prefix file_from_template
+# @class_parameter destination
+# This bundle will define a class file_from_template_${destination}_{kept,repaired,not_ok,ok,reached}
+
+bundle agent file_from_string_mustache(template, destination)
+{
+  vars:
+    "class_prefix"        string => canonify("file_from_template_${destination}");
```
jooooooon: Shouldn't this be file_from_string_mustache?

peckpeck: Maybe, i don't really know what it "should" be

jooooooon: Well unless you have a specific reason for it not to be, this should be the bundle name, so yeah, "file_from_string_mustache"

peckpeck:
PR updated

ncharles:
Why targeting v0.x for 3.7 only code ?
v0.X is used on Rudder 3.1, which uses 3.6 version for the agent

jooooooon:
ncf is not only used by Rudder :-)

ncharles:
who would use branch v0.x with a newer cfengine ?

peckpeck:
PR replaced by https://github.com/Normation/ncf/pull/322

PR #316, opened 2016-02-02
Fixes #7864: Add new checks for special file types

matya:
Check for File being FIFO/Socket/Symlink

### DIFF
--- a/tree/30_generic_methods/_file_check_special.cf
+++ b/tree/30_generic_methods/_file_check_special.cf
@@ -1,0 +1,56 @@
+#####################################################################################
+# Copyright 2013 Normation SAS
+# Copyright 2016 Janos Mattyasovszky
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


jooooooon:
This is a very promising contribution, thanks @matya!

In the interest of staying DRY, it would be preferable to create a 4th generic method, that takes an additional parameter "type", so that all logic is in that bundle, and it just compares the type with the type passed by parameter. That way, the 3 methods you have added here can just be wrappers for this inner method, and none of the logic is duplicated.

See file_copy_from_local_source for a simple example.

matya:
Will check it out tomorrow

matya:
Ok took me quite some time to figure out what you were doing using the new feature of the not always defined this.callers_promisers and copying old and new classes together... Stick not sure I understood all of it, might came back with some questions...

jooooooon:
Ah, yes, that is some confusing s**t. I would recommend targeting your PR to the 0.x branch for now, and not worrying about that stuff (it's only on master)

matya:
retargeting to v0.x

matya:
@jooooooon See https://github.com/Normation/ncf/pull/318

PR #318, opened 2016-02-03
Fixes #7864: Add new checks for special file types

matya:
Check for File being a
- Block device
- Character device
- FIFO/pipe
- Regular file
- Hardlink to an another file
- Socket
- Symlink

### DIFF
--- a/tests/acceptance/30_generic_methods/file_check_FIFO_pipe.cf
+++ b/tests/acceptance/30_generic_methods/file_check_FIFO_pipe.cf
@@ -1,0 +1,89 @@
+#########################################
+#
+# Test checking if a file is a FIFO or not 
+# 
+#########################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "tmp"            string => getenv("TEMP", 1024);
+    "file"           string => "${tmp}/test";
+    "file_canon"     string => canonify("${file}");
+
+    "FIFO_pipe"        string => "${tmp}/FIFO_pipe";
+    "FIFO_pipe_canon"  string => canonify("${FIFO_pipe}");
+
+    "file_list"       slist => { "${file}", "${FIFO_pipe}" };
+
+  files:
+    "${file}"
+      create => "true";
+
+  commands:
+    "/usr/bin/mkfifo"
+      args => "-m 640 \"${FIFO_pipe}\"";
+}
+
+#######################################################
+
+bundle agent test
+{
+  methods:
+    "fce" usebundle => file_check_FIFO_pipe("${init.file_list}");
+}
+
+#######################################################
+
+bundle agent check
+{
+  vars:
+    # Check file type
+    "test_file_type"          string => "/usr/bin/file -b \"${init.file}\"";
+
+    # Check FIFO_pipe type
+    "test_FIFO_pipe_type"     string => "/usr/bin/file -b \"${init.FIFO_pipe}\"";
+
+  classes:
+    "test_file_is_FIFO_pipe"
+      expression => strcmp("fifo (named pipe)", execresult("${test_file_type}", "noshell") ),
+      ifvarclass => "file_check_FIFO_pipe_${init.file_canon}_reached";
+
+    "test_FIFO_pipe_is_FIFO_pipe"
+      expression => strcmp("fifo (named pipe)", execresult("${test_FIFO_pipe_type}", "noshell") ),
+      ifvarclass => "file_check_FIFO_pipe_${init.FIFO_pipe_canon}_reached";
+
+    "ok"   expression => "test_FIFO_pipe_is_FIFO_pipe.file_check_FIFO_pipe_${init.FIFO_pipe_canon}_ok.!test_file_is_FIFO_pipe.!file_check_FIFO_pipe_${init.file_canon}_ok.file_check_FIFO_pipe_${init.file_canon}_reached";
+
+  reports:
+    ok::
+      "$(this.promise_filename) Pass";
+    !ok::
+      "$(this.promise_filename) FAIL";
+    !test_FIFO_pipe_is_FIFO_pipe::
+      "${init.FIFO_pipe} is not a FIFO_pipe as expected. It is '${test_FIFO_pipe_type}'";
+    test_file_is_FIFO_pipe::
+      "${init.file} is a FIFO_pipe, not as expected. It is '${test_file_type}'";
+
+    cfengine::
+      "Check of ${init.file} is not reached"
+        ifvarclass => "!file_check_FIFO_pipe_${init.file_canon}_reached";
+      "Check of ${init.FIFO_pipe} is not reached"
+        ifvarclass => "!file_check_FIFO_pipe_${init.FIFO_pipe_canon}_reached";
+
+}
--- a/tests/acceptance/30_generic_methods/file_check_hardlink.cf
+++ b/tests/acceptance/30_generic_methods/file_check_hardlink.cf
@@ -1,0 +1,98 @@
+#########################################
+#
+# Test checking if a file is hardlinked to an other or not 
+# 
+#########################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "tmp"            string => getenv("TEMP", 1024);
+
+    "file"           string => "${tmp}/test";
+    "file_canon"     string => canonify("${file}");
+
+    "file2"          string => "${tmp}/test2";
+    "file2_canon"    string => canonify("${file2}");
+
+    "hardlink"        string => "${tmp}/hardlink";
+    "hardlink_canon"  string => canonify("${hardlink}");
+
+    "file_list"       slist => { "${file}", "${hardlink}" };
+
+  files:
+    "${file}"
+      create => "true";
+
+    "${file2}"
+      create => "true";
+
+    "${hardlink}"
+      link_from => linkfrom("${file}", "hardlink");
+
+}
+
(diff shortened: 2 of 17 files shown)


tree/30_generic_methods/_file_check_special.cf
```diff
@@ -0,0 +1,56 @@
...
+# @class_parameter file_name 
+# This bundle will define a class file_check_special_${file_type}_${file_name}_{ok, reached, kept} if the
+# file is a special file of type ${file_type} , or file_check_special_${file_type}_${file_name}_{not_ok, reached, not_kept, failed} if
+# the file is not a special file or does not exist
+
+bundle agent file_check_special(file_name, file_type)
+{
+  vars:
+      "file_has_type"       string => filestat("${file_name}", "type");
+      "class_prefix"        string => canonify("file_check_special_${file_type}_${file_name}");
```
jooooooon: This does not match the expected format for ncf bundles. The format is defined by the variables above: class_prefix + class_parameter, so this should be ```"file_check_special_${file_name}"```. Otherwise many things will not behave correctly.

tree/30_generic_methods/file_check_block_device.cf
```diff
@@ -0,0 +1,37 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is block device
+# @description Checks if a file exists and is a block device
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_block_device
```
jooooooon: Since this bundle doesn't do any result class definition itself, this class_prefix should exactly match the class_prefix of the underlying method, file_check_special.

tree/30_generic_methods/file_check_character_device.cf
```diff
@@ -0,0 +1,37 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is character device
+# @description Checks if a file exists and is a character device
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_character_device
```
jooooooon: Same here

tree/30_generic_methods/file_check_fifo.cf
```diff
@@ -0,0 +1,38 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is FIFO/Pipe
+# @description Checks if a file exists and is a FIFO/Pipe
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_FIFO_pipe
```
jooooooon: Same here

tree/30_generic_methods/file_check_regular.cf
```diff
@@ -0,0 +1,38 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is regular
+# @description Checks if a file exists and is a regular file
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_regular_file
```
jooooooon: Same here

tree/30_generic_methods/file_check_socket.cf
```diff
@@ -0,0 +1,37 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is socket
+# @description Checks if a file exists and is a socket
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_socket
```
jooooooon: Same her

tree/30_generic_methods/file_check_symlink.cf
```diff
@@ -0,0 +1,37 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is symlink
+# @description Checks if a file exists and is a symlink
+#
+# @parameter file_name File name
+# 
+# @class_prefix file_check_special_symlink
```
jooooooon: Same here

tree/30_generic_methods/file_check_same.cf
```diff
@@ -0,0 +1,63 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is same
+# @description Checks if two files are the same (hard links)
```
jooooooon: I think this method name is a bit confusing. I originally understood it to mean "files have the same content" and was expecting to see the "diff" command in use.

Maybe we could rename "file_check_hard_links"?

jooooooon:
This is exactly what I had in mind from my comment in #316. Thanks for considering it and adapting your work. This is great stuff!

tree/30_generic_methods/_file_check_special.cf
matya: Well, how would you then check for a file being a socket or being a symlink? The resulting class would be the same for both checks since the file being checked is also the same...

matya: Is it also a problem if this does actually not show up in the editor but is only used as a backend for the others? those work because the ${file_type} is before their ${file_name}...

jooooooon: Ah, no, indeed, probably not! Nice. I hadn't realised what you'd done, I think it's actually just fine as is :)

jooooooon: Checks for class_parameter should not be enforced on "sub-methods" (those that being with _), so this should work.

The best way to check would be to write an acceptance test, but I hardly dare ask you to do more work :)

matya: Will see what I can do

matya: ping @jooooooon, including standardization of resulting class, and also acceptance tests.
`./30_generic_methods/file_check_block_device.cf Pass
./30_generic_methods/file_check_character_device.cf Pass
./30_generic_methods/file_check_exists.cf Pass
./30_generic_methods/file_check_FIFO_pipe.cf Pass
./30_generic_methods/file_check_hardlink.cf Pass
./30_generic_methods/file_check_regular.cf Pass
./30_generic_methods/file_check_socket.cf Pass
./30_generic_methods/file_check_symlink.cf Pass
./30_generic_methods/file_check_symlinkto.cf Pass
`

tests/acceptance/30_generic_methods/file_check_FIFO_pipe.cf
```diff
@@ -0,0 +1,90 @@
...
+
+    "FIFO_pipe"        string => "${tmp}/FIFO_pipe";
+    "FIFO_pipe_canon"  string => canonify("${FIFO_pipe}");
+
+    "file_list"       slist => { "${file}", "${FIFO_pipe}" };
+
+  files:
+    "${file}"
+      create => "true",
+      perms  => mog("000", "root", "0");
```
jooooooon: You should probably just remove this "perms" line, it is setting *no* permissions (I assume it was copy pasted from a different acceptance test...). Using "000" for perms may break this test for unrelated reasons.

Omitting the perms body should just use sane defaults.

matya: @jooooooon done. I also made the mknod/mkfifo create the device with 640.

tests/acceptance/30_generic_methods/file_check_FIFO_pipe.cf
```diff
@@ -0,0 +1,89 @@
...
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("$#this.promise_filename}") };
```
jooooooon: This looks like a typo: ```$#this```

And it seems to be in several test files!

matya: I take it over from a different check, however, the check seems to work... :-)

matya: Have opened https://www.rudder-project.org/redmine/issues/7907 for it, will change PR on the weekend.

matya: @jooooooon it's fixed now.

tree/30_generic_methods/_file_check_special.cf
```diff
@@ -0,0 +1,56 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File check is special
+# @description Checks if a file exists and is a special file
+#
+# @parameter file_name File name
+# @parameter file_type File type
```
peckpeck: Since this methods can be directly used by users, the type parameter should have a list of accepted values

jooooooon:
@matya All that's missing to merge this PR is a little bit of extra doc (list of accepted parameters). Can you add these in, so this can be merged, or should we do it?

tests/acceptance/30_generic_methods/file_check_hardlink.cf
```diff
@@ -0,0 +1,98 @@
...
+    "negative"  usebundle => file_check_hardlink("${init.file2}", "${init.hardlink}");
+}
+
+#######################################################
+
+bundle agent check
+{
+  vars:
+    # Check file type
+    "test_file_devino"        string => "/bin/stat -c %d:%i \"${init.file}\"";
```
ncharles: ha - on all distrib I have with me, the stat path is /usr/bin/stat

matya: Well, on SLES it is symlinked, so usr/bin would also work.

```shell
linux:~ # ll /usr/bin/stat
lrwxrwxrwx 1 root root 9 Dec  8 22:29 /usr/bin/stat -> /bin/stat

linux:~ # which stat
/bin/stat
```

ncharles: Ha, interesting, my SLES is opposite for which
> orchestrateur-3:~ # ls -al /usr/bin/stat
> lrwxrwxrwx 1 root root 9 23 févr.  2015 /usr/bin/stat -> /bin/stat
> orchestrateur-3:~ # which stat
> /usr/bin/stat

ncharles:
PR replaced by https://github.com/Normation/ncf/pull/341

tests/acceptance/30_generic_methods/file_check_hardlink.cf
matya: Yeah, well, it depends on what you PATH is. :-) Mine is:

```shell
# echo $PATH
/sbin:/usr/sbin:/usr/local/sbin:/bin:/usr/bin:<...>
```

ncharles:
Replaced by https://github.com/Normation/ncf/pull/342

PR #319, opened 2016-02-04
Fixes #7875: Broken tests after #7838

amousset:
https://www.rudder-project.org/redmine/issues/7875


amousset:
Backported the optionnal tag feature to 0.x.

PR #320, opened 2016-02-07
Fixes #7882: make curl follow redirects

matya:
Example where API gives you a redirect is when the authentication info is given in the URL and after authentication you get forwarded.

```
Q: ".../bin/echo "'{"a":   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Q: ".../bin/echo "'{"a":                                  Dload  Upload   Total   Spent    Left  Speed
124   249  124   249    0     0  64291      0 --:--:-- --:--:-- --:--:-- 83000
Q: ".../bin/echo "'{"a": <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
Q: ".../bin/echo "'{"a": <html><head>
Q: ".../bin/echo "'{"a": <title>302 Found</title>
Q: ".../bin/echo "'{"a": </head><body>
Q: ".../bin/echo "'{"a": <h1>Found</h1>
Q: ".../bin/echo "'{"a": <p>The document has moved <a href="/tinker/check_mk/login.py?_origtarget=webapi.py?action%3dadd_host">here</a>.</p>
Q: ".../bin/echo "'{"a": </body></html>
```

### DIFF
--- a/tree/30_generic_methods/http_request_check_status_headers.cf
+++ b/tree/30_generic_methods/http_request_check_status_headers.cf
@@ -49,7 +49,7 @@ bundle agent http_request_check_status_headers(method, url, expected_status, hea
 
   commands:
       "${paths.path[curl]}"
-        args => "-I -s -o /dev/null -w \"%{http_code}\" ${url} -X ${method} -H '${headers}' | /bin/grep -q ${expected_status}",
+        args => "-I -L -s -o /dev/null -w \"%{http_code}\" ${url} -X ${method} -H '${headers}' | /bin/grep -q ${expected_status}",
         contain => in_shell,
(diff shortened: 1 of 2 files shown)


jooooooon:
Looks good! There is also a curl call in "http_request_check_status_headers", could you add it there too?

jooooooon:
@matya Any chance you could add in that missing option?

jooooooon:
Also, can you please rename the commit so it starts with "Fixes #7882:" please?

matya:
@jooooooon done, hope it's not a problem if it's in two separate commits, I could merge both if @FlorianHeigl would not mind :)

jooooooon:
That's fine. Thanks @matya.

PR #326, opened 2016-02-19
Fixes #7947: Add a generic methods doc output for Rudder

amousset:
https://www.rudder-project.org/redmine/issues/7947

### DIFF
--- a/tools/ncf_doc_rudder.py
+++ b/tools/ncf_doc_rudder.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Usage: ./ncf_docrudder.py
+#
+# This is a Python module to generate documentation from generic methods in ncf to be embedded in Rudder manual
+
+import ncf 
+
+from pprint import pprint
+
+def rudder_version(version):
+  # CFEngine <=> Rudder versions
+
+  # some generic methods have ">= version", others ">=version"
+  canonified_version = version.replace(" ", "")
+  
+  cfengine_version = {}
+  cfengine_version[">=3.5"] = "2.10"
+  cfengine_version[">=3.6"] = "2.11"
+  cfengine_version[">=3.7"] = "3.2"
+    
+  if canonified_version in cfengine_version:
+    return cfengine_version[canonified_version]
+  else:
+    return "unknown"
+
+if __name__ == '__main__':
+
+  # Get all generic methods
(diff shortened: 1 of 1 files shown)


tools/ncf_doc_rudder.py
```diff
@@ -0,0 +1,69 @@
...
+def rudder_version(version):
+  # CFEngine <=> Rudder versions
+
+  # some generic methods have ">= version", others ">=version"
+  canonified_version = version.replace(" ", "")
+  
+  cfengine_version = {}
+  cfengine_version[">=3.5"] = "2.10"
+  cfengine_version[">=3.6"] = "2.11"
+  cfengine_version[">=3.7"] = "3.2"
```
amousset: Do you know if there is a better source for this information somewhere?

amousset:
The idea is to generate a markdown file targeted for Rudder doc integration.

rudder-doc will generate it, convert it to asciidoc using pandoc, and integrate it in the reference section of the manual.

tools/ncf_doc_rudder.py
peckpeck: It should be in rudder-info and if it's not, we should add it

tools/ncf_doc_rudder.py
```diff
@@ -0,0 +1,69 @@
...
+# This is a Python module to generate documentation from generic methods in ncf to be embedded in Rudder manual
+
+import ncf 
+
+from pprint import pprint
+
+def rudder_version(version):
+  # CFEngine <=> Rudder versions
+
+  # some generic methods have ">= version", others ">=version"
```
peckpeck: And some may have <version some day

tools/ncf_doc_rudder.py
amousset: How can I add it?

jooooooon:
I like the concept of this. However, I think it belongs in the rudder-doc repo - because almost everything in this file is specific to the rudder-doc.

If one day we refactor Rudder documentation (change title levels, change format altogether, etc), that will be a in a big pull request on rudder-doc, and it would make sense for this file to be changed at the same time.

I see it that rudder-doc has dependencies on other data sources (in this case, the ncf source code), and it should pull it in where necessary.

Does that make sense or am I missing a bigger point here?

amousset:
The first idea was only to do an asciidoc output, but now it has Rudder-specific stuff, it should indeed go to rudder-doc repo.

PR #336, opened 2016-03-31
Fixes #8137: Technique editor display method params of the last generic method created for all methods

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8137


ncharles:
Thank you @VinceMacBuche !

PR #339, opened 2016-04-07
Fixes #8160: Remote file copies in ncf should be encrypted

amousset:
https://www.rudder-project.org/redmine/issues/8160


ncharles:
Awesome, thank you so much

PR #341, opened 2016-04-18
Bug 7864/missing functions for file checks

ncharles:


jooooooon:
This should be targeted to v0.x branch

PR #342, opened 2016-04-19
Bug 7864/missing functions for file checks

ncharles:


ncharles:
replacing https://github.com/Normation/ncf/pull/341

PR #343, opened 2016-04-19
Fixes #8194: file_check_ on branch master use deprecated _logger interface

ncharles:
https://www.rudder-project.org/redmine/issues/8194

### DIFF
--- a/tree/30_generic_methods/_file_check_special.cf
+++ b/tree/30_generic_methods/_file_check_special.cf
@@ -33,7 +33,12 @@ bundle agent file_check_special(file_name, file_type)
 {
   vars:
       "file_has_type"       string => filestat("${file_name}", "type");
-      "class_prefix"        string => canonify("file_check_${file_type}_${file_name}");
+
+      "old_class_prefix"    string => canonify("file_check_${file_type}_${file_name}");
+      "promisers"            slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
+      "class_prefix"        string => canonify(join("_", "promisers"));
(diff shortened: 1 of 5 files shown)


tree/30_generic_methods/_file_check_special.cf
```diff
@@ -44,13 +49,17 @@ bundle agent file_check_special(file_name, file_type)
   methods:
     exists_special::
       "file_exists_and_is_special"
+        usebundle => _classes_success("${old_class_prefix}");
+      "file_exists_and_is_special"
         usebundle => _classes_success("${class_prefix}");
```
jooooooon: We would normally only call the _classes_success method for the new class_prefix if the has_promiser_stack class is defined. See for example https://github.com/Normation/ncf/blob/master/tree/30_generic_methods/_service_check_running_src.cf#L50.

tree/30_generic_methods/_file_check_special.cf
```diff
@@ -44,13 +49,17 @@ bundle agent file_check_special(file_name, file_type)
...
 
     !exists_special::
       "file_doesnt_exists_or_is_not_special"
+        usebundle => _classes_failure("${old_class_prefix}");
+      "file_doesnt_exists_or_is_not_special"
         usebundle => _classes_failure("${class_prefix}");
```
jooooooon: Same here

jooooooon:
This looks good, aside from the one query about protecting use of the new class_prefix with ```has_promiser_stack```. Maybe @peckpeck can tell us how important this is?

jooooooon:
@peckpeck confirmed the _class_{success,failure} methods don't care

PR #350, opened 2016-05-08
Fixes #8228: Add constraints to ncf parameters

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8228

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -16,6 +16,8 @@ import shutil
 import sys
 import os
 import codecs
+import ncf_constraints
+from pprint import pprint
 
 # Additionnal path to look for cf-promises
 additional_path = ["/opt/rudder/bin","/usr/sbin","/usr/local"]
@@ -27,10 +29,9 @@ dirs = [ "10_ncf_internals", "20_cfe_basics", "30_generic_methods", "40_it_ops_k
 
 tags = {}
 tags["common"] = ["bundle_name", "bundle_args"]
-tags["generic_method"] = ["name", "description", "documentation", "parameter", "class_prefix", "class_parameter", "class_parameter_id", "agent_version"]
+tags["generic_method"] = ["name", "description", "documentation", "parameter", "class_prefix", "class_parameter", "class_parameter_id", "agent_version", "parameter_constraint"]
 tags["technique"] = ["name", "description", "version"]
-
-optionnal_tags = [ "documentation" ]
+optionnal_tags = [ "documentation", "parameter_constraint" ]
 multiline_tags = [ "description", "documentation" ]
 
 class NcfError(Exception):
@@ -137,7 +138,10 @@ def parse_generic_method_metadata(technique_content):
 
 def parse_bundlefile_metadata(content, bundle_type):
   res = {}
+  warnings = []
   parameters = []
+  param_names = set() 
+  param_constraints = {} 
   multiline = False
   previous_tag = None
 
@@ -153,12 +157,19 @@ def parse_bundlefile_metadata(content, bundle_type):
       if tag in tags[bundle_type]:
         # tag "parameter" may be multi-valued
         if tag == "parameter":
-          parameters.append({'name': match.group(3), 'description': match.group(4)})
+          param_name = match.group(3)
+          parameters.append({'name': param_name, 'description': match.group(4)})
+          param_names.add(param_name)
+        if tag == "parameter_constraint":
+          constraint = json.loads("{" + match.group(4)+ "}")
+          param_constraints.setdefault(match.group(3), {}).update(constraint)
         else:
           res[tag] = match.group(2)
         previous_tag = tag
         continue
 
+    
+     
     # Parse line without tag, if previous tag was a multiline tag
     if previous_tag is not None and previous_tag in multiline_tags:
       match = re.match("^\s*# (.*)$", line, flags=re.UNICODE)
@@ -178,7 +189,6 @@ def parse_bundlefile_metadata(content, bundle_type):
 
     # read a complete bundle definition
     match = re.match("[^#]*bundle\s+agent\s+(\w+)(\(([^)]+)\))?[^(]*$", match_line, flags=re.UNICODE|re.MULTILINE|re.DOTALL)
-    #match = re.match("[^#]*bundle\s+agent\s+(\w+)(\(([^)]+)\))?.*$", line, flags=re.UNICODE)
     if match:
       multiline = False
       res['bundle_name'] = match.group(1)
@@ -199,8 +209,24 @@ def parse_bundlefile_metadata(content, bundle_type):
       name = res['bundle_name'] if 'bundle_name' in res else "unknown"
       raise NcfError("The class_parameter name \"" + res['class_parameter'] + "\" does not seem to match any of the bundle's parameters in " + name)
 
+  # Check that we don't have a constraint that is defined on a non existing parameter:
+  wrong_constraint_names = set(param_constraints.keys()) - param_names
+  if len(wrong_constraint_names) > 0:
+      warning_message = "In technique '' defining constraint on non existing parameters: "+ ", ".join(wrong_constraint_names)
+      print(warning_message)
+      warnings.append(warning_message)
+ 
   # If we found any parameters, store them in the res object
   if len(parameters) > 0:
+    for param in parameters:
+      parameter_name = param["name"]
+      constraints = param_constraints.get(param["name"], ncf_constraints.default_constraint)
+      for key, constraint in constraints.iteritems():
+        check = ncf_constraints.check_constraint_type(key, constraint)
+        if not check['result']:
+          raise NcfError("Value for constraint '" + key + "' of parameter '"+ param['name'] +"' is not valid, "+", ".join(check["errors"]))
+      param["constraints"] = constraints
+    
     res['parameter'] = parameters
     
   if bundle_type == "generic_method" and not "agent_version" in res:
@@ -213,7 +239,8 @@ def parse_bundlefile_metadata(content, bundle_type):
     name = res['bundle_name'] if 'bundle_name' in res else "unknown"
     raise NcfError("One or more metadata tags not found before the bundle agent declaration (" + ", ".join(missing_keys) + ") in " + name)
 
-  return res
+  result = { "result" : res, "warnings" : warnings }
+  return result
 
 
 def class_context_and(a, b):
@@ -434,7 +461,7 @@ def canonify_class_context(class_context):
   return regex.sub(r'",canonify("\1"),"', class_context)
 
 
-def generate_technique_content(technique_metadata):
+def generate_technique_content(technique_metadata, methods):
   """Generate technique CFEngine file as string from its metadata"""
 
   technique = add_default_values_technique_metadata(technique_metadata)
@@ -451,9 +478,17 @@ def generate_technique_content(technique_metadata):
 
   # Handle method calls
   for method_call in technique["method_calls"]:
+    method_name = method_call["method_name"]
     regex = re.compile(r'(?<!\\)"', flags=re.UNICODE )
     # Treat each argument of the method_call
     if 'args' in method_call:
+      for index, arg in enumerate(method_call['args']):
+        parameter = methods[method_name]["parameter"][index]
+        arg_constraint = parameter.get("constraints", {})
+        check = ncf_constraints.check_parameter(arg, arg_constraint)
+        if not check['result']:
+            error_constraint = "'"+"', '".join(check['errors'])+"'"
+            raise NcfError("Invalid value for parameter '"+ parameter["name"] +"' of method '"+ method_name +"' in technique '"+ technique['bundle_name'] +"', invalid value is '"+ arg+"', doesn't match constraints: " + error_constraint)
       args = ['"%s"'%regex.sub(r'\\"', arg) for arg in method_call['args'] ]
       arg_value = ', '.join(args)
     else:
@@ -481,12 +516,15 @@ def get_all_techniques_metadata(include_methods_calls = True, alt_path = ''):
 
   filenames = get_all_techniques_filenames(alt_path)
   errors = []
+  warnings = []
 
   for file in filenames:
     with codecs.open(file, encoding="utf-8") as fd:
       content = fd.read()
     try:
-      metadata = parse_technique_metadata(content)
+      result = parse_technique_metadata(content)
+      metadata = result["result"]
+      warnings.extend(result["warnings"])
       all_metadata[metadata['bundle_name']] = metadata
 
       if include_methods_calls:
@@ -497,7 +535,7 @@ def get_all_techniques_metadata(include_methods_calls = True, alt_path = ''):
       errors.append(error)
       continue # skip this file, it doesn't have the right tags in - yuk!
 
-  return { "data": all_metadata, "errors": format_errors(errors) }
+  return { "data": { "techniques" : all_metadata }, "errors": format_errors(errors), "warnings": warnings }
 
 
 def get_all_generic_methods_metadata(alt_path = ''):
@@ -505,19 +543,22 @@ def get_all_generic_methods_metadata(alt_path = ''):
 
   filenames = get_all_generic_methods_filenames(alt_path)
   errors = []
+  warnings = []
 
   for file in filenames:
     with codecs.open(file, encoding="utf-8") as fd:
       content = fd.read()
     try:
-      metadata = parse_generic_method_metadata(content)
+      result = parse_generic_method_metadata(content)
+      metadata = result["result"]
+      warnings.extend(result["warnings"])
       all_metadata[metadata['bundle_name']] = metadata
     except NcfError as e:
       error = NcfError("Could not parse generic method in '" + file + "'", cause=e )
       errors.append(error)
       continue # skip this file, it doesn't have the right tags in - yuk!
 
-  return { "data": all_metadata, "errors": format_errors(errors) }
+  return { "data": { "generic_methods" : all_metadata }, "errors": format_errors(errors), "warnings": warnings }
 
 
 def write_technique(technique_metadata, alt_path = ""):
@@ -528,7 +569,7 @@ def write_technique(technique_metadata, alt_path = ""):
     path = alt_path
 
   try:
-    
+    methods= get_all_generic_methods_metadata(alt_path)["data"]["generic_methods"]    
     # Check if file exists
     bundle_name = technique_metadata['bundle_name']
     filename = os.path.realpath(os.path.join(path, "50_techniques", bundle_name, bundle_name+".cf"))
@@ -547,7 +588,7 @@ def write_technique(technique_metadata, alt_path = ""):
     execute_hooks("pre", action, path, bundle_name)
 
     # Write technique file
-    content = generate_technique_content(technique_metadata)
+    content = generate_technique_content(technique_metadata, methods)
     with codecs.open(filename, "w", encoding="utf-8") as fd:
       fd.write(content)
 
--- a/tools/ncf_constraints.py
+++ b/tools/ncf_constraints.py
@@ -1,0 +1,147 @@
+import re
+import subprocess
+import json
+import os.path
+import shutil
+import sys
+import os
+import codecs
+from pprint import pprint
+
+if sys.version_info < (3,):
+  string_type = unicode
+else:
+  string_type = str
+
+### Constraint checking function
+
+# Helper function on constraint check return
+def return_constraint(result, error_message):
+  if result:
+    return None
+  else:
+    return error_message
+
+def select( parameter_value, accepted_result):
+  result = parameter_value in accepted_result
+  error_message = "Value not in accepted values: "+ str(accepted_result)
+  return return_constraint(result, error_message)
+
+def max_length( parameter_value, max_size):
+  current_length = len(parameter_value)
+  result = current_length <= max_size
+  error_message = "Max length is " + str(max_size) + ". Current size is " + str(current_length)
+  return return_constraint(result, error_message)
+
+def min_length( parameter_value, min_size):
+  current_length = len(parameter_value)
+  result = current_length >= min_size
+  error_message = "Min length is " + str(min_size) + ". Current size is " + str(current_length)
+  return return_constraint(result, error_message)
+
+def match_regex(parameter_value, regex):
+  match = re.search(regex, parameter_value, re.U)
+  result = match is not None
+  error_message = "Must match regex "+ regex
+  return return_constraint(result, error_message)
+
+def not_match_regex(parameter_value, regex):
+  result = match_regex(parameter_value, regex) is not None
+  error_message = "Must not match regex "+ regex
+  return return_constraint(result, error_message)
+
+def allow_empty_string(parameter_value, allow):
+  # If length more that one, this is an error
+  result = allow or min_length(parameter_value,1) is None
+  error_message = "Must not be empty"
+  return return_constraint(result, error_message)
+
+def allow_whitespace_string(parameter_value, allow):
+  # If leading/trailing whitespace is not allowed, and we have some, this is an error
+  leading = match_regex(parameter_value, r'^\s') is None
+  trailing = match_regex(parameter_value, r'\s$') is None
+  result = allow or ( not leading and not trailing )
+  error_message = "Must not have leading or trailing whitespaces"
+  return return_constraint(result, error_message)
+
+
+constraints = {
+    "select" : {
+        "check" : select
+      , "type"  : list
+    }
+  , "allow_whitespace_string" : {
+        "check" : allow_whitespace_string
+      , "type"  : bool
+    }
+  , "allow_empty_string" : {
+        "check" : allow_empty_string
+      , "type"  : bool
+    }
+  # use unicode for regexs, since they will be parsed as unicode ...
+  , "regex" :  {
+        "check" : match_regex
+      , "type"  : string_type
+    }
+  , "not_regex":  {
+        "check" : not_match_regex
+      , "type"  : string_type
+    }
+  , "max_length" :  {
+        "check" : max_length
+      , "type"  : int
+    }
+  , "min_length" :  {
+        "check" : min_length
+      , "type"  : int
+    }
+}
+
+variable_constraints = {
+  "allow_whitespace_string" : constraints["allow_whitespace_string"]
+}
+
+default_constraint = {
+    "allow_whitespace_string" : False
+  , "allow_empty_string" : False
+}
+
+def check_parameter(parameter_value, parameter_constraints):
+  """Checks that a parameter value is ok with the constraint of the parameter"""
+
+  result = True
+  errors = []
+  constraint_set = constraints
+
+  # Check that our value contains a variable or not
+  if re.search(r'[$@][{(][$@{(a-zA-Z0-9[\]_.-]+[})]', parameter_value) is not None :
+    constraint_set = variable_constraints
+
+  for (constraint_name, constraint_value) in parameter_constraints.iteritems():
+    if constraint_name in constraint_set:
+      constraint = constraint_set[constraint_name]
+      constraint_check = constraint['check'](parameter_value,constraint_value)
+      if constraint_check is not None:
+        result = False
+        errors.append(constraint_check)
+
+  check = {'result': result, 'errors': errors}
+  return check
+
+def check_constraint_type(key, value):
+
+  result = True
+  errors = []
+
+  if key in constraints:
+    constraint = constraints[key]
+    result = type(value) is constraint["type"]
+    if not result:
+      errors.append("expected a value of type '" + constraint["type"].__name__ + "', got '"+ str(value) +"' of type '"+ type(value).__name__  + "'")
+  else:
+    result = False
+    errors.append("'"+ key +"' is an unknwown constraint")
+  check = {'result': result, 'errors': errors}
+  return check
+
+
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -236,7 +236,7 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location, 
     $http.get('/ncf/api/techniques',data).
       success(function(data, status, headers, config) {
 
-        $.each( data.data, function(techniqueName, technique_raw) {
+        $.each( data.data.techniques, function(techniqueName, technique_raw) {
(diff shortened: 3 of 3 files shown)


tools/ncf.py
```diff
@@ -16,6 +16,8 @@
 import sys
 import os
 import codecs
+import ncf_constraints
+from pprint import pprint
```
jooooooon: You can remove this now, I guess it was just used for debugging

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  return parameter in accepted_result
+
+def max_length( parameter, max_size):
+  return parameter.length <= max_size
+
+def min_length( parameter, min_size):
+  return parameter.length >= min_size
+
+def match_regexp(parameter, regexp):
+  match = re.match(regexp, parameter, re.S)
```
jooooooon: Why have you used the ```re.S``` option here?

I know that is causes the regexp expression ```.``` to match newlines like any other character. What brought you to choose to do this here? I think we should at least write a comment to explain, and certainly document it somewhere.

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+
+constraint_type = {
+    "allow_whitespace_string" : bool
+  , "allow_empty_string" : bool
+  , "list" : list
+  # use unicode for regexps, since they will be parsed as unicode ... 
+  , "regexp" : unicode
+  , "not_regexp": unicode
+  , "max" : int
+  , "min" : int
```
jooooooon: Please rename max and min max_length and min_length.

We probably want to keep max and min available for actual maximum values (on an int, for example)

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  string_regex = string_constraints[string_constraint]
+  result = match_regexp(parameter, string_regex)
+
+  # Check if in list of accepted values
+  if result:
+    if "list" in constraint:
+      result = from_list(parameter, constraint["list"])
+
+  # Check regexps constraints
+  if result:
```
jooooooon: You know you could delete this line and the others like it, right? :)

jooooooon: Oh, OK, no you couldn't. I see what you're doing here.

tools/ncf.py
```diff
@@ -201,6 +206,15 @@ def parse_bundlefile_metadata(content, bundle_type):
 
   # If we found any parameters, store them in the res object
   if len(parameters) > 0:
+    for param in parameters:
+      constraints = {}
+      if len(param_constraints) > 0:
+        constraints = dict( (k,v) for constraint in param_constraints if constraint["name"] == param["name"] for (k,v) in constraint["constraint"].iteritems() )
+      for key, constraint in constraints.iteritems():
+        if not ncf_constraints.check_constraint_type(key, constraint):
+          raise NcfError("Value for parameter constraint '" + key + "' is not valid, expected a value of type '" + ncf_constraints.constraint_type[key].__name__ + "', got '"+ constraint+"' of type '"+ type(constraint).__name__  + "'")
```
jooooooon: This error message should say which ```param``` this constraint error is for!

tools/ncf_constraints.py
jooooooon: I actually think we should implement this a bit differently. It's hard to explain, so let me propose a PR to your PR branch.

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  return result
+
+
+constraint_type = {
+    "allow_whitespace_string" : bool
+  , "allow_empty_string" : bool
+  , "list" : list
+  # use unicode for regexps, since they will be parsed as unicode ... 
+  , "regexp" : unicode
+  , "not_regexp": unicode
```
jooooooon: regexP is a pretty uncommon name. We usually just call them "regex".

tools/ncf_constraints.py
jooooooon: Also, this should be ```re.search``` instead, see https://docs.python.org/2/library/re.html#search-vs-match (re.match only matches at the beginning of a string)

tools/ncf_constraints.py
jooooooon: Done, see https://github.com/VinceMacBuche/ncf/pull/1

tools/ncf.py
```diff
@@ -201,6 +206,15 @@ def parse_bundlefile_metadata(content, bundle_type):
 
   # If we found any parameters, store them in the res object
   if len(parameters) > 0:
+    for param in parameters:
+      constraints = {}
+      if len(param_constraints) > 0:
+        constraints = dict( (k,v) for constraint in param_constraints if constraint["name"] == param["name"] for (k,v) in constraint["constraint"].iteritems() )
```
peckpeck: can't this be split for a better readability ?

tools/ncf.py
```diff
@@ -154,6 +156,9 @@ def parse_bundlefile_metadata(content, bundle_type):
         # tag "parameter" may be multi-valued
         if tag == "parameter":
           parameters.append({'name': match.group(3), 'description': match.group(4)})
+        if tag == "parameter_constraint":
+          constraint = json.loads("{" + match.group(4)+ "}")
+          param_constraints.append({'name': match.group(3), 'constraint': constraint})
```
peckpeck: wouldn't it be easier if you used a dict name=>constraints for param_constraints ? 
Or do you allow multiple constraint line for a single parameter ?

tools/ncf.py
```diff
@@ -154,6 +156,9 @@ def parse_bundlefile_metadata(content, bundle_type):
         # tag "parameter" may be multi-valued
         if tag == "parameter":
           parameters.append({'name': match.group(3), 'description': match.group(4)})
+        if tag == "parameter_constraint":
+          constraint = json.loads("{" + match.group(4)+ "}")
```
peckpeck: Do you expect a list of key values or a list of elements here ? 
Did you write the constraint format somewhere ?

tools/ncf.py
VinceMacBuche: We want to allow multiple line for a single parameter :(

tools/ncf.py
VinceMacBuche: a list a key:values 

Not documented yet ... shoudl add it in doc / ncf.io

tools/ncf.py
```diff
@@ -451,9 +465,16 @@ def generate_technique_content(technique_metadata):
...
   # Handle method calls
   for method_call in technique["method_calls"]:
+    method_name = method_call["method_name"]
     regex = re.compile(r'(?<!\\)"', flags=re.UNICODE )
     # Treat each argument of the method_call
     if 'args' in method_call:
+      for index, arg in enumerate(method_call['args']):
+        parameter = methods[method_name]["parameter"][index]
+        arg_constraint = parameter.get("constraints", {})
+        check = ncf_constraints.check_parameter(arg, arg_constraint)
```
peckpeck: What if we have a variable indirection in the argument ?

VinceMacBuche: What do you mean ? that we don't have the same number or order in arguments and spec ?

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  return check
+
+
+constraint_type = {
+    "allow_whitespace_string" : bool
+  , "allow_empty_string" : bool
+  , "list" : list
+  # use unicode for regexs, since they will be parsed as unicode ...
+  , "regex" : unicode
+  , "not_regex": unicode
```
peckpeck: unicode is not valid in python3, it should be str instead

tools/ncf.py
peckpeck: Let's say my constraint is "regex=^\d+$" and my argument is "0${var}"
the constraint check will fail but my argument is valid.

tools/ncf_constraints.py
VinceMacBuche: str does not work for unicode 'strings' in python 2 :'(

VinceMacBuche: need to find a way to make it work maybe an array of type and check wether one is fine

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  return not match_regex(parameter, regex)
+
+def check_parameter(parameter, constraint):
+  """Checks that a parameter is ok with the constraint passed as parameter"""
+
+  result = True
+  errors = [];
+
+  # First check 'string' constraints
+  if not constraint.get("allow_empty_string", False):
```
peckpeck: You should add comments on the default implied by the way you get the constraint.
Here: default allow empty string

tools/ncf_constraints.py
peckpeck: won't work, unicode doesn't exist at all in python 3.
You must detect the version and do string=str or string=unicode somewhere

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+    if match_regex(parameter, "^$"):
+      errors.append("allow_empty_string")
+      result = False
+
+  if not constraint.get("allow_whitespace_string", False):
+    if match_regex(parameter, r'^\s') or match_regex(parameter, r'\s+$'):
```
peckpeck: From what I understand "allow_whitespace_string" allow string containing only whitespaces.
then match_regex(parameter, r'^\s+$') is enough.

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+      errors.append("allow_whitespace_string")
+      result = False
+
+  # Check if in list of accepted values
+  if "list" in constraint:
+    if not from_list(parameter, constraint["list"]):
```
peckpeck: Why did you define sub functions for a single line here ?

tools/ncf_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+  if "min_length" in constraint:
+    if not min_length(parameter, constraint["min_length"]):
+      errors.append("min_length")
+      result = False
+
+  if "max_length" in constraint:
+    if not max_length(parameter, constraint["max_length"]):
+      errors.append("max_length")
+      result = False
+
```
peckpeck: What you are doing for the result is a logical AND between all checks.
This is valid from a developer point of view, but what I would expect from a more natural language would be a logical OR for some constraints.

Especially for the list constraint: the list are things that are allowed and other constraints are how to allow things that are not in the list.

tools/ncf_constraints.py
jooooooon: Actually, it's the opposite: default = don't allow empty string :)

peckpeck: Very good case for comments

jooooooon: Very true. I propose:

```# If empty strings are not allowed, and we have one, this is an error```

File: tools/ncf_constraints.py (hunk 13 above)
Comment (jooooooon) [inline tools/ncf_constraints.py]: Nope, it allows there to be whitespace surrounding valid content, such as " packagename    "

Comment (jooooooon) [inline tools/ncf_constraints.py]: A comment would come in here too, like

```# If leading/trailing whitespace is not allowed, and we have some, this is an error```

File: tools/ncf_constraints.py (hunk 15 above)
Comment (jooooooon) [inline tools/ncf_constraints.py]: This touches on the specs for this feature, which were done a while ago now. We thought back then that it would be sufficient for the first implementation to have a list of *required* constraints only. This makes sense for many combinaisons (min_length + max_length + matches regex + forbidden regex, for example).

The list can be useful in conjuction with other constraints to *relax* the strictness: for example ```constraints: 'list': {'a', 'b', 'c'}, 'allow_whitespace_string': true``` would allow values a, b or c, but also tolerate leading/trailing whitespace.

File: tools/ncf.py (hunk 10 above)
Comment (jooooooon) [inline tools/ncf.py]: Wow, good catch. That's a very good point!

Should we implement the same behaviour as Rudder-web and replace "${.*}" and "$(.*)" with ".*" in the regex, do you think?

File: tools/ncf_constraints.py (hunk 13 above)
Comment (peckpeck) [inline tools/ncf_constraints.py]: Then there is still a bug since you call match and not search

Comment (peckpeck) [inline tools/ncf_constraints.py]: I don't like pure english sentence here since you need to fully read it to understand it.
I'd prefer something like "Default: leading/trailing whitespace forbiden"

Comment (jooooooon) [inline tools/ncf_constraints.py]: The ```match_regex``` function is a local wrapper for ```re.search``` (not match, which is a Python weirdness and should be abolished, IMHO :) )

File: tools/ncf_constraints.py (hunk 15 above)
Comment (peckpeck) [inline tools/ncf_constraints.py]: There have been not release with those specs, so they can still change.

Adding other constraints with a logical AND to a list is not really useful, allowing the trailing whitespace case is not a constraint but a filter (trim if needed), a list is already a full set of constraint in itself and adding some can only confirm existing constraints or contradict them.

File: tools/ncf_constraints.py (hunk 13 above)
Comment (peckpeck) [inline tools/ncf_constraints.py]: my previous question, was why wrap single line call with a function ? especially if you change the name ?

File: tools/ncf.py (hunk 10 above)
Comment (peckpeck) [inline tools/ncf.py]: You cannot since it is in the value you should replace the variable, not in the regex.
Since you cannot guess in advance something that will match the regex you have no other solution than canceling accept the input.

Comment (VinceMacBuche) [inline tools/ncf.py]: If there is variable we cannot determine the value and hence it's almost impossible to check any constraint ...
After discussing with @peckpeck, we think that if there is a variable we should always validate the value ...

File: tools/ncf_constraints.py (hunk 15 above)
Comment (VinceMacBuche) [inline tools/ncf_constraints.py]: I think we are overengineering for now ... I suggest we should keep the easy way now with AND (should validate all constraint) 
I'm not sure use cases with OR will happen

**Comment by VinceMacBuche** [comment]
PR updated

**Comment by VinceMacBuche** [comment]
As a reminder: 
Still need to add comments.
Check default cases are still treated (maybe define default constraint: allow_empty : false, allow_witthepace: false)

File: tools/ncf.py (hunk 1 above)
Comment (jooooooon) [inline tools/ncf.py]: Up?

File: tools/ncf_constraints.py (hunk 15 above)
Comment (jooooooon) [inline tools/ncf_constraints.py]: peckpeck wrote:
> There have been not release with those specs, so they can still change.

Whoa! Well, yes, of course we can change them. I'm not trying to start a war here, just trying to explain to you how we got to this implementation.

I don't know why you weren't with us when we discussed the specs, but a summary of them was shared by email weeks ago (April 15th, to be precise). I would have expected to have this discussion at the time. This is a *technical* review, so rather off topic to discuss specs (as a reminder, the goal of a TR is to make sure the implementation works and matches the specs).

However, I am happy to rediscuss the specs if you feel the naming or logic is flawed. As usual, I'd much rather spend more time to make something good than have to support something bad for years.

My understanding is that you're querying two points here:

1) Naming: the main reason for this PR is to allow to trim leading/trailing whitespace and/or whitespace only values. All values are currently automatically trimmed by ncf-builder, and we want to be able to tell ncf-builder to not do this via this feature. I agree that "allowing" things isn't really a *constraint*, more a "relaxation of an implicit constraint". However, we failed to find a better name. Do you have any ideas?

2) The need to have an extra "or" operator as well as the implicit "and" operator.

Did I understand your comments correctly?

One answer in particular:
> Adding other constraints with a logical AND to a list is not really useful, [...] a list is already a full set of constraint in itself and adding some can only confirm existing constraints or contradict them.

I agree, it's not particularly useful to have "list" and other constraints. But I don't see that as a problem - if users don't want to use both, they won't. However, each taken separately *is* useful, therefore I think that makes this a valid feature.

File: tools/ncf_constraints.py [hunk 16]
```diff
@@ -0,0 +1,107 @@
...
+def allow_whitespace_string(parameter, allow):
+  return allow or not_match_regex(parameter, r'^\s') or not_match_regex(parameter, r'\s$')
+
+
+constraints = {
+    "list" : {
+        "check" : from_list
+      , "type"  : list
+    }
```
jooooooon: Re-reading the email discussion we had about specs makes me notice the name you've used here: "list". I don't think this is a good name, because it is not clear what "constraint" "list" is imposing.

I prefer "enum" (which, to me at least, makes it clear that this constraint in an enumeration of possible values), but I'm open to other terms. Maybe "values"?

Thoughts?

tools/ncf.py
ncharles: I think we should best effort. If we have the constraint "regex=^\d+$" but type in foo${bar}, their is no way it will match the regexp.

tools/ncf.py
```diff
@@ -201,6 +206,17 @@ def parse_bundlefile_metadata(content, bundle_type):
 
   # If we found any parameters, store them in the res object
   if len(parameters) > 0:
+    for param in parameters:
+      constraints = {}
+      if len(param_constraints) > 0:
+        constraint_array = [ constraint["constraint"] for constraint in param_constraints if constraint["name"] == param["name"] ]
```
ncharles: shouldn't we warn if a constraint refer to a variable name that is not defined ?

tools/ncf_constraints.py
ncharles: if it's an enumeration, in the web world we use "select", which is fairly easily understood.

tools/ncf.py
peckpeck: I never remove pprint, you always need it.
Having to add it each time you debug and removing it just after is very tiring. Moreover it costs almost nothing.

tools/ncf.py
peckpeck: You would need your own regex interpreter to be able to tell that.

tools/ncf.py
VinceMacBuche: I clearly agree with peck, we always need it

tools/ncf_constraints.py
VinceMacBuche: I like the 'select' ! second choice would be 'values'. I don't like 'enum', and agree that list is not a good choice

tools/ncf.py
VinceMacBuche: I think a warning is a good idea
At least that should not be an 'error' that prevents reading the technique

jooooooon: That's what we agreed upon in the specs dicsussion by email:

> On envoie la méthode sans la contraintes, mais avec un warning associé qui indique ce qui ne va pas (et qu'on retrouve, dans les logs et l'api )

tools/ncf.py
jooooooon: Well, yes and no. The *only* way to actually enforce these constraints with variables would be to check at runtime - that would be a great feature to have, but that's obviously not possible at this stage.

However, we can *try*. A simple approach would be to remove the variable contents and test what's left, something like:

```
re.sub(r'[\$@][\{\(][a-zA-Z0-9\[\]_.-]+[\}\)]', '', arg)
```

However, this could fail if the field contains only a variable, which could then cause a failure is allow_empty_string is not true. Should we therefore automatically add that constraint *if* we had to remove a variable? That seems logical enough to me.

tools/ncf.py
jooooooon: OK, fine. It's nice to answer comments though, even if you don't agree with them :)

tools/ncf_constraints.py
jooooooon: 'select' works for me. It matches nicely with Rudder's Techniques metadata.xml terminology, too.

tools/ncf.py
peckpeck: This won't work and we will often have valid values that will be considered invalid.
It's not even valid for the simple example I gave if you remove the leading zero.

VinceMacBuche: So we discussed in a meeting about this issue with @jooooooon @peckpeck 
In most case, when a variable is involved, checking contraint of it is undecidable and that it can only be checked at runtime (which is not the scope of this fix!) ( impossible for regex constraints, min length, at least very complicated for 'select' )

We decided that: when a variable is detected
* Skip constraint check that cannot be done
* Add an info log that constraint on that value could not be check
* Do all check that can be done (max_length, trailing/leading whitespaces)

@jooooooon @peckpeck , Am i missing something ?

@ncharles (as you reacted on this i ask you the question) and anyone who read that PR: What do you think of it ?

jooooooon: Correct, except for one detail:

> Add an info log that constraint on that value could not be check

Not a *log*, a message in the API response. We don't want our logs filling up with useful information such as "someone used a variable".

tools/ncf_constraints.py
VinceMacBuche: We discussed this with @jooooooon and @peckpeck 

We decided that we should keep the first specs for two reasons: It's easy to understand, and for now we don't know if there is some usecase where we would like to have complex relationship between our constraint.

Nevertheless, if we encounter a usecase, or if you have any to share, we will be happy to open a new issue and discuss of the evolution of that feature

tools/ncf.py
VinceMacBuche: yes info response in the api, thanks @jooooooon to precise this :)

ncharles: ok for me

tools/ncf_constraints.py
```diff
@@ -0,0 +1,107 @@
...
+def check_parameter(parameter, parameter_constraints):
+  """Checks that a parameter is ok with the constraint passed as parameter"""
+
+  result = True
+  errors = []
+
+  for (constraint_name, constraint_value) in parameter_constraints.iteritems():
+    if not constraint_name in constraints:
+      result = False
+      errors.append (constraint_name + " is not a valid constraint")
```
jooooooon: Please don't store *config* errors (from the generic method's implementation) in the same field as *validation* errors (from the user provided values). The "errors" field I created is used to have useful logging output in ncf.py (something like """Value bla doesn't match constraints " ".join("errors")""").

I propose you rename them "failed_constraints" and keep the failed constraints in that one, and use the 'errors' field as a human readable message to log/return via API.

tools/ncf_constraints.py
```diff
@@ -0,0 +1,107 @@
...
+  return match is not None
+
+def not_match_regex(parameter, regex):
+  return not match_regex(parameter, regex)
+
+def allow_empty_string(parameter, allow):
+  return allow or not_match_regex(parameter, "^$")
+
+def allow_whitespace_string(parameter, allow):
+  return allow or not_match_regex(parameter, r'^\s') or not_match_regex(parameter, r'\s$')
```
jooooooon: This does not do what you want it to: it will return "True" on the first value in the "or"-connected list that is True, therefore it succeeds (wrongly) on values like "bla " (because there is no space at the beginning)

It should be:

```
return allow or (not_match_regex(parameter, r'^\s') and not_match_regex(parameter, r'\s$'))
```

VinceMacBuche: Exactly! Nice catch

Le 13 mai 2016 17:20:49 GMT+02:00, Jonathan Clarke <notifications@github.com> a écrit :
>> +
>> +def min_length( parameter, min_size):
>> +  return len(parameter) >= min_size
>> +
>> +def match_regex(parameter, regex):
>> +  match = re.search(regex, parameter)
>> +  return match is not None
>> +
>> +def not_match_regex(parameter, regex):
>> +  return not match_regex(parameter, regex)
>> +
>> +def allow_empty_string(parameter, allow):
>> +  return allow or not_match_regex(parameter, "^$")
>> +
>> +def allow_whitespace_string(parameter, allow):
>> +  return allow or not_match_regex(parameter, r'^\s') or
>not_match_regex(parameter, r'\s$')
>
>This does not do what you want it to: it will return "True" on the
>first value in the "or"-connected list that is True, therefore it
>succeeds (wrongly) on values like "bla " (because there is no space at
>the beginning)
>
>It should be:
>
>```
>return allow or (not_match_regex(parameter, r'^\s') and
>not_match_regex(parameter, r'\s$'))
>```
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Normation/ncf/pull/350/files/76b1c2546686678e1db499012c21a5911f5e1dc8#r63201268

-- 
Envoyé de mon téléphone Android avec K-9 Mail. Excusez la brièveté.

VinceMacBuche:
PR updated

tools/ncf.py
peckpeck: The variable case is still missing

peckpeck:
There should be an issue for documenting the constraint syntax

peckpeck:
This PR needs a rebase

tools/ncf_constraints.py
```diff
@@ -0,0 +1,111 @@
...
+
+def max_length( parameter, max_size):
+  return len(parameter) <= max_size
+
+def min_length( parameter, min_size):
+  return len(parameter) >= min_size
+
+def match_regex(parameter, regex):
+  match = re.search(regex, parameter)
+  return match is not None
```
peckpeck: re.S and re.M should be considered
It means that a parameter should be considered as a whole and we are matching it as is, not with separate lines. A new line can still be matched with \n, which is expected. And we can match things that couldn't be matched without those flags. 

Example:
- the regex r"^A.*\nB.*$" matches a line starting with A followed by a line starting with B. This is not possible without re.M
- with re.S r"A.*B.*" could match 2 lines with A in the first one and B in the second one. This may be counter intuitive. 

So I'd say, use re.MULTILINE only.

Anyway, whatever the final behavior, the multiline case must be carefully documented.

peckpeck: We should use re.UNICODE too, because all our strings are unicode and i would expect '\w' to match 'é' in the general case.

tools/ncf_constraints.py
```diff
@@ -0,0 +1,111 @@
...
+
+def not_match_regex(parameter, regex):
+  return not match_regex(parameter, regex)
+
+def allow_empty_string(parameter, allow):
+  # If empty strings are not allowed, and we have one, this is an error
+  return allow or not_match_regex(parameter, "^$")
```
peckpeck: Perfect example, without re.M, this would match anything containing an empty line and not just an empty string.

peckpeck:
Everything else is ok for me

tools/ncf_constraints.py
VinceMacBuche: I don't think we should use re.M

I think we want to check the whole value with our regex, so with you first case you would match 
    A
    A
    B

If you use re.M and would not match with other cases, and i think we don't want to match

peckpeck: You're right, I mixed up the definition of re.M, it is the opposite.

VinceMacBuche:
@peckpeck I made a separate comment so you can look on the changes

tools/ncf_constraints.py
```diff
@@ -14,34 +15,59 @@
...
-  match = re.match(regex, parameter_value, re.U)
-  return match is not None
+  match = re.search(regex, parameter_value, re.U)
+  result = match is not None
+  error_message = "Must match regex "+ regex
+  return return_constraint(result, error_message)
 
 def not_match_regex(parameter_value, regex):
-  return not match_regex(parameter_value, regex)
+  result = match_regex(parameter_value, regex) is not None
```
peckpeck: There should be a re.U here

tools/ncf_constraints.py
```diff
@@ -14,34 +15,59 @@
...
-  return allow or min_length(parameter_value,1)
+  # If length more that one, this is an error
+  result = allow or min_length(parameter_value,1) is None
+  error_message = "Must not be empty"
+  return return_constraint(result, error_message)
 
 def allow_whitespace_string(parameter_value, allow):
   # If leading/trailing whitespace is not allowed, and we have some, this is an error
-  return allow or ( not_match_regex(parameter_value, r'^\s') and not_match_regex(parameter_value, r'.*\s$') )
+  not_leading = not_match_regex(parameter_value, r'^\s') is None
```
peckpeck: not x = (not fct == not error) 

Too many negations here, couldn't you do it with a more straightforward expression ?

tools/ncf_constraints.py
```diff
@@ -14,34 +15,59 @@
...
+  not_trailing = not_match_regex(parameter_value, r'\s$') is None
+  result = allow or ( not_leading and not_trailing )
+  error_message = "Must not have leading or trailing whitespaces"
```
peckpeck: Where is the doc for this ? 
Because I think "allow whitespace" is a bit different from "allow leading and trailing whitespace"

tools/ncf_constraints.py
```diff
@@ -71,21 +71,33 @@ def allow_whitespace_string(parameter, allow):
...
-  """Checks that a parameter is ok with the constraint passed as parameter"""
+def check_parameter(parameter_value, parameter_constraints):
+  """Checks that a parameter value is ok with the constraint of the parameter"""
 
   result = True
   errors = []
+  constraint_set = constraints
+
+  # Check that our value contains a variable or not
+  value_without_variables = re.sub(r'[\$@][\{\(][a-zA-Z0-9\[\]_.-]+[\}\)]', "", parameter_value)
```
peckpeck: all backslaslash are unnecessary here, except for the \]

tools/ncf_constraints.py
VinceMacBuche: It is changed in next commit :)

VinceMacBuche: and match_regexp is not re.match !

tools/ncf_constraints.py
peckpeck: This won't match cfengine indexed arrays "${value[${index}]}"
To have a real variable parsing you need an expression parser, since we don't want an exact variable matching, but just know that a variable in present, you can just add ${( in the allowed character within the name

tools/ncf_constraints.py
peckpeck: Yes, but if it is not obvious it must be documented in an obvious place.

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

tools/ncf_constraints.py
```diff
@@ -114,11 +114,8 @@ def check_parameter(parameter_value, parameter_constraints):
   constraint_set = constraints
 
   # Check that our value contains a variable or not
-  value_without_variables = re.sub(r'[\$@][\{\(][a-zA-Z0-9\[\]_.-]+[\}\)]', "", parameter_value)
-  if parameter_value != value_without_variables:
+  if re.search(r'[$@][{(][$@{(a-zA-Z0-9[\]_.-]+[})]', parameter_value) is not None :
```
peckpeck: Just for information, ][ instead of [\] works too

VinceMacBuche:
PR updated

PR #352, opened 2016-05-09
Fixes #8109: Missing reporting in Permissions (non recursive) when the file does not exist

jooooooon:
https://www.rudder-project.org/redmine/issues/8109


peckpeck:
OK, merging

tree/30_generic_methods/permissions_type_recursion.cf
```diff
@@ -96,6 +97,11 @@ bundle agent permissions_type_recursion(path, mode, owner, group, type, recursio
         classes      => classes_generic("${class_prefix}");
 
   methods:
+
+    !target_exists::
```
VinceMacBuche: Does the report is still sent if the class is defined ? (my cfengine fu tells me that report will only be done if !target_exists ... ) I think an "any::" should be added after this usebundle

jooooooon: Absolutely! Great catch, thanks @VinceMacBuche 

Fixed in #353

PR #353, opened 2016-05-11
Fixes #8266: Missing reporting in Permissions (non recursive) when the file does not exist - missing class in original fix

jooooooon:
https://www.rudder-project.org/redmine/issues/8266


peckpeck:
Good catch
Merging

PR #356, opened 2016-05-16
Fixes #8305: ncf tests don't run on master branch due to old CFEngine version - missing apt-https module

jooooooon:
https://www.rudder-project.org/redmine/issues/8305


peckpeck:
Good, merging

PR #357, opened 2016-05-17
Fixes #8095: add generic method to manage key -> value , without array

ncharles:
https://www.rudder-project.org/redmine/issues/8095

### DIFF
--- a/tests/acceptance/30_generic_methods/file_ensure_key_value.cf
+++ b/tests/acceptance/30_generic_methods/file_ensure_key_value.cf
@@ -1,0 +1,154 @@
+#######################################################
+#
+# Test checking if a key-value pair are present in a file 
+#
+#######################################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "tmp"         string => getenv("TEMP", 1024);
+    "file1"       string => "${tmp}/test1";
+    "ref_file1"   string => "${tmp}/ref1";
+    "file_canon1" string => canonify("${file1}");
+
+    "file2"       string => "${tmp}/test2";
+    "ref_file2"   string => "${tmp}/ref2";
+    "file_canon2" string => canonify("${file2}");
+
+    "file3"       string => "${tmp}/test3";
+    "ref_file3"   string => "${tmp}/ref3";
+    "file_canon3" string => canonify("${file3}");
+
+
+# First test: ensure that a value is correctly modified
+    "base_text1"  string => "# File with lhs=rhs text
+JAVA_OPTS=\"-Djava.awt.headless=true -Xmx128m -XX:+UseConcMarkSweepGC\"";
+
+    "ref_text1"   string=> "# File with lhs=rhs text
+JAVA_OPTS=\\\"-Djava.awt.headless=true -Xmx1024m\\\"";
+
+    "key1"        string => "JAVA_OPTS";
+    "value1"      string => "\"-Djava.awt.headless=true -Xmx1024m\"";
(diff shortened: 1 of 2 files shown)


tree/30_generic_methods/file_ensure_key_value.cf
```diff
@@ -0,0 +1,47 @@
...
+  files:
+      "${file}"
+        create        => "true",
+        edit_line     => maintain_key_values("file_ensure_key_value.keys", "${separator}"),
+        edit_defaults => ncf_empty_select("false"),
+        classes       => classes_generic("${class_prefix}");
+
+  methods:
+      "report"
+        usebundle  => _logger("Ensure line in format key${separator}value into ${file}", "${class_prefix}");
```
jooooooon: s/into/in/

tests/acceptance/30_generic_methods/file_ensure_key_value.cf
```diff
@@ -0,0 +1,154 @@
...
+    "file2_correct"
+       expression => returnszero("${file_diff_test2}", "noshell"),
+       ifvarclass => canonify("file_ensure_key_value_${init.file_canon2}_reached");
+    "file3_correct"
+       expression => returnszero("${file_diff_test3}", "noshell"),
+       ifvarclass => canonify("file_ensure_key_value_${init.file_canon3}_reached");
+
+
+    "ok_test1"     expression => "file1_correct.file_ensure_key_value_${init.file_canon1}_repaired.!file_ensure_key_value_${init.file_canon1}_not_ok";
+    "ok_test2"     expression => "file2_correct.file_ensure_key_value_${init.file_canon2}_repaired.!file_ensure_key_value_${init.file_canon2}_not_ok";
```
jooooooon: Can you add ".!file_ensure_key_value_${init.file_canon2}_kept" to these two tests please?

tree/30_generic_methods/file_ensure_key_value.cf
```diff
@@ -0,0 +1,47 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File ensure key -> value present
+# @description Ensure that the file contains a pair of "key separator value", with arbitrary separator between each key and its value. It the key is already present, it will change the value associated to this key.
```
jooooooon: This is rather a long description. Could you split it out into a single line summary in @description and a more detailed explanation in @documentation? See https://github.com/Normation/ncf/blob/master/tree/30_generic_methods/variable_dict.cf#L20 for an example.

jooooooon: And http://www.ncf.io/pages/reference.html#variable_dict to see what it looks like in the end

tests/acceptance/30_generic_methods/file_ensure_key_value.cf
ncharles: no, because file already exists, so promise is kept

jooooooon: That is *exactly* what I was concerned about. What you're saying right there is CFEngine pseudo-logic, and not at all logic to the rest of the human species.

In ncf, we aim to have simple, clean methods: file_ensure_key_value should be repaired in the file needed changes, and kept if the file was already OK, nothing else.

Can you look into a workaround to avoid this confusing and nasty kept class? If it comes from the "create => true" part of the promise, maybe just separate this into two promises, and report only on the second? If not, how about a checksum before and after the file edit promise, and if the checksum differs, then force the repaired class, if not force the kept class?

ncharles:
Commit modified

tests/acceptance/30_generic_methods/file_ensure_key_value.cf
ncharles: this is a bug present in all file edition generic method, so it needs to be handled in another ticket to fix all editions

ncharles: http://www.rudder-project.org/redmine/issues/8320

jooooooon: OK, good point. I opened the ticket here, for future reference: http://www.rudder-project.org/redmine/issues/8321

PR #361, opened 2016-05-19
Fixes #8339: Tests are broken on 1.x branch because of old logger being used

amousset:
https://www.rudder-project.org/redmine/issues/8339


peckpeck:
Good, merging

PR #362, opened 2016-05-19
Fixes #8341: Missing promise types in ordering check script

amousset:
https://www.rudder-project.org/redmine/issues/8341


jooooooon:
👍

PR #363, opened 2016-05-19
Fixes #8343: logger_rudder breaks if expected_reports.csv.res is missing but expected_reports.csv.tmp is present

jooooooon:
https://www.rudder-project.org/redmine/issues/8343

### DIFF
--- a/tree/30_generic_methods/logger_rudder.cf
+++ b/tree/30_generic_methods/logger_rudder.cf
@@ -33,14 +33,14 @@ bundle agent logger_rudder(message, class_prefix)
       "expected_reports_temp"   string => "${expected_reports_source}.tmp";
       "expected_reports_file"   string => "${expected_reports_source}.res";
 
-    logger_rudder_temp_resfile_repaired::
+    (!final_resfile_exists.logger_rudder_temp_resfile_reached|logger_rudder_temp_resfile_repaired)::
       # 2/ If the temporary file has been updated, read it to get all values in an array, and canonify the first entry
       "dim" int => readstringarrayidx("reports", "${expected_reports_temp}", "\s*#[^\n]*", ";;", 9999, 999999);
       "keys" slist => getindices("reports");
 
       "reports[${keys}][canon]" string => canonify("${reports[${keys}][1]}");
 
-    (logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept)::
+    (logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept.!logger_rudder_temp_resfile_repaired.final_resfile_exists)::
       "c_class_prefix" string => canonify("${class_prefix}");
 
       # 4/ Once the final expected reports file has been expanded, read in our array
@@ -51,6 +51,8 @@ bundle agent logger_rudder(message, class_prefix)
       "report_data_read" expression => isgreaterthan("${number_lines}","0");
 
       "keys_defined"     expression => isvariable("keys");
+
+      "final_resfile_exists" expression => fileexists("${expected_reports_file}");
 
   files:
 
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/logger_rudder.cf
```diff
@@ -33,14 +33,14 @@ bundle agent logger_rudder(message, class_prefix)
...
+    (!final_resfile_exists.logger_rudder_temp_resfile_reached|logger_rudder_temp_resfile_repaired)::
       # 2/ If the temporary file has been updated, read it to get all values in an array, and canonify the first entry
       "dim" int => readstringarrayidx("reports", "${expected_reports_temp}", "\s*#[^\n]*", ";;", 9999, 999999);
       "keys" slist => getindices("reports");
 
       "reports[${keys}][canon]" string => canonify("${reports[${keys}][1]}");
 
-    (logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept)::
+    (logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept.!logger_rudder_temp_resfile_repaired.final_resfile_exists)::
       "c_class_prefix" string => canonify("${class_prefix}");
```
ncharles: I don't understand this condition.
If resfile was repaired, or ( tmp file was correct and res file exists ) do the reporting
So in case when resfile was correct but tmp file was not (like, it was removed), we won't have the first report.

I think the condition should be:
(logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept.final_resfile_exists)::

jooooooon: One of the challenges here is not entering this "block" too early - we need to only get into on the third pass (pass 1: check if the temporary file needs to be updated, pass 2: read keys from temporary file + write the final expected reports). Therefore we must make sure other promises have been evaluated first (see below).

Put simply, we need to enter this block if and only if we know we've successfully obtained the final expected reports file. This can happen for one of two reasons: **1)** it needing fixing and was fixed (easy: ```logger_rudder_final_resfile_repaired```) or **2)** it didn't need fixing but it is OK (as in we've already checked it).

The second reason is harder to get a class for. The ```logger_rudder_final_resfile_kept``` is not enough, because it may not be defined, since it is a conditional block (if ```final_resfile_exists``` and ```!logger_rudder_temp_resfile_repaired```). In effect, you could therefore say either we have the ```logger_rudder_final_resfile_kept``` class or we have the condition that means that promise was entirely skipped: ```logger_rudder_final_resfile_kept|(final_resfile_exists.!logger_rudder_temp_resfile_repaired)```.

Your example (in case when resfile was correct but tmp file was not) would lead to ```logger_rudder_temp_resfile_repaired```, and therefore the ```logger_rudder_final_resfile_kept``` class would be set, because that promise would be reached.

Therefore we should be able to use:

```
logger_rudder_final_resfile_repaired|logger_rudder_final_resfile_kept|(final_resfile_exists.!logger_rudder_temp_resfile_repaired)
```

BUT this last condition (```(final_resfile_exists.!logger_rudder_temp_resfile_repaired)```) can cause this block to be entered on pass 2, too early. I suggest we simply add in a pass1/pass2/pass3 set of classes and use:

```
logger_rudder_final_resfile_repaired|logger_rudder_final_resfile_kept|(pass3.final_resfile_exists.!logger_rudder_temp_resfile_repaired)
```

WDYT?

ncharles: this is super complex :/
but this looks correct. merging

jooooooon: Er, you merged but I hadn't actually change the code. I was just asking your opinion. 
-- 
Sent from Kaiten Mail. Please excuse my brevity.

jooooooon:
@ncharles ? Did you change your mind?

ncharles:
Ha - I was rechecking all the conditions, and what you had in the promises seemed to match the same cases as the complex expression you suggested. At least in my mind of a Friday afternoon.
Let me recheck

ncharles:
Ok, I've been cross checking and it works as expected in the following cases:
- it tmp file doesn't exist or is wrong, and res file doesn't exist, is invalid or is valid
- if tmp file is correct, and res file doesn't exists or is valid

If tmp file is correct, and res file is incorrect but exists, the res file will stay incorrect (as before) - but this is expected by construction of this bundle

PR #367, opened 2016-05-23
Fixes #6488: Avoid waiting on lock when calling yum

ncharles:
https://www.rudder-project.org/redmine/issues/6488


tree/20_cfe_basics/cfengine/packages.cf
```diff
@@ -99,6 +99,19 @@ bundle common rpm_knowledge
       "rpm3_arch_regex" string => "\S+\s+(\S+).*";
 }
 
+bundle common redhat_no_locking_knowledge {
+  vars:
+    !redhat_setopt_not_available::
+      "no_locking_command" string => "--setopt=exit_on_lock=True";
+    redhat_setopt_not_available::
+      "no_locking_command" string => "";
```
jooooooon: I know this is nitpicking, but it's really not a *command*, it's an *option*.

tree/20_cfe_basics/cfengine/packages.cf
```diff
@@ -99,6 +99,19 @@ bundle common rpm_knowledge
       "rpm3_arch_regex" string => "\S+\s+(\S+).*";
 }
 
+bundle common redhat_no_locking_knowledge {
+  vars:
+    !redhat_setopt_not_available::
```
jooooooon: This class is sensitive to normal ordering. In my tests, the result of the real test below didn't matter, this case was always set. I strongly recommend adding a "check_done" class below and prefixing it to this condition.

tree/20_cfe_basics/cfengine/packages.cf
```diff
@@ -99,6 +99,19 @@ bundle common rpm_knowledge
...
+      "no_locking_command" string => "--setopt=exit_on_lock=True";
+    redhat_setopt_not_available::
+      "no_locking_command" string => "";
+
+
+  classes:
+    redhat::
+      "redhat_setopt_not_available" expression => returnszero("LANG=C ${paths.path[yum]} --setopt=exit_on_lock=True 2>&1 | ${paths.path[grep]} -E '(Command line error: no such option: --setopt|Main config did not have a exit_on_lock attr. before setopt)'", "useshell");
```
jooooooon: This returns a false positive when yum is simply not available.

I propose reversing the test:

` "redhat_setopt_available" not => returnszero("LANG=C ${paths.path[yum]} --setopt=exit_on_lock=True 2>&1 | ${paths.path[grep]} -E '(Command line error: no such option: --setopt|Main config did not have a exit_on_lock attr. before setopt)'", "useshell"); `

tree/20_cfe_basics/cfengine/packages.cf
ncharles: classes in bundle common are global - so i'm not super happy to add extra classes there that will leak in global context and might overlap

tree/20_cfe_basics/cfengine/packages.cf
ncharles: this option is only used in conjonction with yum binary, and checked on redhat only.
I agree that reversing the condition will make it clearer

ncharles:
Commit modified

jooooooon:
Looks good to me, merging. We should also propose this fix upstream.

PR #369, opened 2016-05-25
Fixes #8379: invalid reporting in ncf

jooooooon:
https://www.rudder-project.org/redmine/issues/8379

### DIFF
--- a/tree/30_generic_methods/logger_rudder.cf
+++ b/tree/30_generic_methods/logger_rudder.cf
@@ -48,8 +48,10 @@ bundle agent logger_rudder(message, class_prefix)
         int        => getfields("^${current_technique_report_info.technique_name};;${c_class_prefix};;.*", "${expected_reports_file}", ";;", "report_data");
 
   classes:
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
Good catch !!

tree/30_generic_methods/logger_rudder.cf
```diff
@@ -48,8 +48,10 @@ bundle agent logger_rudder(message, class_prefix)
         int        => getfields("^${current_technique_report_info.technique_name};;${c_class_prefix};;.*", "${expected_reports_file}", ";;", "report_data");
 
   classes:
+    (logger_rudder_final_resfile_repaired|logger_rudder_temp_resfile_kept.!logger_rudder_temp_resfile_repaired.final_resfile_exists)::
```
ncharles: I don't understand why it changes anything: these classes are defined globally the first time the bundle is run, and issue was at last call of this bundle
I testes this, and it repairs the issue, but it really make no sense ...

ncharles: ok, it does make sense, because od the final_resfile_exists being defined locally

PR #371, opened 2016-05-31
Fixes #8430: logger_rudder can re-read expected_reports_temp even without change, impacting performance significantly

jooooooon:
https://www.rudder-project.org/redmine/issues/8430

### DIFF
--- a/tree/30_generic_methods/logger_rudder.cf
+++ b/tree/30_generic_methods/logger_rudder.cf
@@ -33,7 +33,7 @@ bundle agent logger_rudder(message, class_prefix)
       "expected_reports_temp"   string => "${expected_reports_source}.tmp";
       "expected_reports_file"   string => "${expected_reports_source}.res";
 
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/logger_rudder.cf
```diff
@@ -33,7 +33,7 @@ bundle agent logger_rudder(message, class_prefix)
       "expected_reports_temp"   string => "${expected_reports_source}.tmp";
       "expected_reports_file"   string => "${expected_reports_source}.res";
 
-    (!final_resfile_exists.logger_rudder_temp_resfile_reached|logger_rudder_temp_resfile_repaired)::
+    pass2.(!final_resfile_exists.logger_rudder_temp_resfile_reached|logger_rudder_temp_resfile_repaired)::
```
ncharles: ha, of course. 
final_resfile_exists is not yet defined, but all other classes are already defined (classes as promise outcome remain for the rest of the run)
good catch

PR #372, opened 2016-06-01
Fixes #8433: logger_rudder evaluates too late content of temp file, causing it to miss first report

ncharles:
https://www.rudder-project.org/redmine/issues/8433


peckpeck:
OK, merging

PR #374, opened 2016-06-14
Fixes #8522: Doc does not generated after 8228

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8522

### DIFF
--- a/tests/unit/test_ncf.py
+++ b/tests/unit/test_ncf.py
@@ -16,8 +16,9 @@ class TestNcf(unittest.TestCase):
       self.technique_content = fd.read()
     with open(self.test_generic_method_file) as fd:
       self.generic_method_content = fd.read()
-    
-    self.technique_metadata = ncf.parse_technique_metadata(self.technique_content)
+    self.all_methods = ncf.get_all_generic_methods_metadata()["data"]["generic_methods"]
+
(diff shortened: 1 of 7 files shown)


tests/unit/test_ncf.py
```diff
@@ -16,8 +16,9 @@ def setUp(self):
       self.technique_content = fd.read()
     with open(self.test_generic_method_file) as fd:
       self.generic_method_content = fd.read()
-    
-    self.technique_metadata = ncf.parse_technique_metadata(self.technique_content)
+    self.all_methods = ncf.get_all_generic_methods_metadata()["data"]["generic_methods"]
```
peckpeck: Why did you add this new variable here ?

tests/unit/test_ncf.py
```diff
@@ -303,7 +304,7 @@ def test_delete_technique(self):
 
   def test_generate_technique_content(self):
     """Check that generate_technique_content works correctly - including escaping " in arguments"""
-    content = ncf.generate_technique_content(self.technique_metadata_test)
+    content = ncf.generate_technique_content(self.technique_metadata_test, self.all_methods)
```
peckpeck: Ah ok, here

PR #380, opened 2016-06-16
Fixes #5659: Make list_compatible_inputs script equivalent for Windows

ncharles:
https://www.rudder-project.org/redmine/issues/5659


ncharles:
Commit modified

PR #390, opened 2016-06-22
Fixes #8323: Limit the length of technique names

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8323

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -65,11 +65,8 @@
               <div class="form-group">
                 <label for="cloneName" class="control-label">New Technique name:</label>
                 <div>
-                  <input  type="text" id="cloneName" name="name" techniquename class="form-control" ng-model="technique.name" placeholder="Technique Name" required ng-maxlength="2048">
+                  <input  type="text" id="cloneName" name="name" techniquename class="form-control" ng-model="technique.name" placeholder="Technique Name" required>
                 </div>
-                <span class="text-danger" ng-show="cloneForm.name.$error.maxlength">
-                  Fields over 2048 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
-                </span>
                 <span class="text-danger pull-left" ng-show="cloneForm.name.$error.techniqueName">Technique name must be unique</span>
               </div>
             </div>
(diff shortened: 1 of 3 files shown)


builder/index.html
```diff
@@ -282,33 +276,27 @@ <h3 >{{getMethodName(selectedMethod)}}
...
-                                     Fields over 2048 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
+                                  <input  type="text" name="cfClasses" id="advanced"  class="form-control" ng-change="updateClassContext()" ng-model="selectedMethod.advanced_class" placeholder="">
+                                </div>
+                                <div class="form-group condition-form">
+                                  <label for="class_context">CFEngine class context:</label>
+                                  <input ng-readonly="true" type="text" id="" name="cfClassContext" class="form-control" ng-model="selectedMethod.class_context" style="cursor:text" ng-maxlength="2048">
+                                  <span class="text-danger" ng-show="CForm.form.cfClassContext.$error.maxlength">
+                                    Fields over 2048 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
```
VinceMacBuche: Should be:

'Classes over 2048 characters are currently not supported.'

builder/index.html
```diff
@@ -328,10 +316,7 @@ <h5 class="control-label method-param" >{{parameter.name}}:
                                   </h5>
                                 </label>
                                 <br style="clear:both"/>
-                                <textarea msd-elastic name="param" class="form-control" rows="1" ng-required="true" id="param-{{$index}}" ng-model="parameter.value" placeholder="" ng-maxlength="2048"></textarea>
-                                <span class="text-danger" ng-show="paramForm.param.$error.maxlength">
-                                  Fields over 2048 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
-                                </span>
+                                <textarea msd-elastic name="param" class="form-control" rows="1" ng-required="true" id="param-{{$index}}" ng-model="parameter.value" placeholder=""></textarea>
```
VinceMacBuche: It's ok to keep this one !!

RaphaelGauthier:
PR updated

builder/index.html
```diff
@@ -170,27 +167,24 @@ <h3 >{{originalTechnique.name}} <span ng-hide="originalTechnique.name">New Techn
...
-                                </span>
                             </div>
                             <div class="form-group" show-errors>
                                 <label for="bundleName" class="col-sm-3 control-label">Bundle name:</label>
                                 <div class="col-sm-8">
-                                    <input disabled id="bundleName" bundlename name="bundle_name" class="form-control" ng-model="selectedTechnique.bundle_name">
+                                    <input disabled id="bundleName" bundlename name="bundle_name" class="form-control" ng-model="selectedTechnique.bundle_name" ng-maxlength="255">
                                 </div>
+                                <span class="text-danger col-sm-8 col-sm-offset-3" ng-show="editForm.bundle_name.$error.maxlength">
+                                    Fields over 255 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
```
VinceMacBuche: You can replace text here. Don't know what yet, but it can't be that message

RaphaelGauthier:
PR updated

builder/index.html
```diff
@@ -65,11 +65,8 @@ <h3 class="modal-title">Clone Technique</h3>
               <div class="form-group">
                 <label for="cloneName" class="control-label">New Technique name:</label>
                 <div>
-                  <input  type="text" id="cloneName" name="name" techniquename class="form-control" ng-model="technique.name" placeholder="Technique Name" required ng-maxlength="2048">
+                  <input  type="text" id="cloneName" name="name" techniquename class="form-control" ng-model="technique.name" placeholder="Technique Name" required>
                 </div>
-                <span class="text-danger" ng-show="cloneForm.name.$error.maxlength">
-                  Fields over 2048 characters are currently not supported. Please insert your content into a file, and copy it with a file_copy_* method, or use a template.
```
ncharles: why are you removing the 2048 character limit ?

VinceMacBuche: Name has no limit it's the bundle name that has one, and it will be checked on the form after validating that popup

RaphaelGauthier: Because the Technique name is only used in the application so it does not require length check.
On the other hand, this check has to be done on the Technique bundle name which is used in the URL.

PR #391, opened 2016-06-23
Fixes #7971: Non applicable reports "Not applicable is not applicable" are useless and can be duplicates, breaking reporting

jooooooon:
https://www.rudder-project.org/redmine/issues/7971

### DIFF
--- a/tests/unit/test_ncf_rudder.py
+++ b/tests/unit/test_ncf_rudder.py
@@ -87,15 +87,15 @@ class TestNcfRudder(unittest.TestCase):
     expected_result.append('')
     expected_result.append('    !(debian)::')
     expected_result.append('      "dummy_report" usebundle => _classes_noop("service_start_apache2");')
-    expected_result.append('      "dummy_report" usebundle => logger_rudder("Not applicable", "service_start_apache2");')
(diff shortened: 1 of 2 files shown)


tests/unit/test_ncf_rudder.py
```diff
@@ -125,6 +138,8 @@ def test_rudder_reporting_with_variable_content(self):
 
     generated_result = ncf_rudder.generate_rudder_reporting(self.reporting_with_var_metadata)
 
+    print(self._unidiff_output(result, generated_result))
```
ncharles: there is an extra print there

jooooooon:
Commit modified

tests/unit/test_ncf_rudder.py
```diff
@@ -125,6 +138,11 @@ def test_rudder_reporting_with_variable_content(self):
 
     generated_result = ncf_rudder.generate_rudder_reporting(self.reporting_with_var_metadata)
 
+    diff = self._unidiff_output(result, generated_result)
```
VinceMacBuche: Maybe you can add a comment so we knwo why do that (failing test output are impossible to understand otherwise)

VinceMacBuche:
Appart the comment I suggest to add on the diff, you can add ready to merge label

jooooooon:
Commit modified

PR #397, opened 2016-07-07
Fixes #8637: Failing unit tests - because of #8610

amousset:
https://www.rudder-project.org/redmine/issues/8637


amousset:
Commit modified

PR #402, opened 2016-07-08
Fixes #8660: Adapt documetation url of each method to us rudder documentation if editor is used with Rudder

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8660


fanf:
We need to have that pr in NCF 0.X to, else people not migrated will have a broken doc.

VinceMacBuche:
PR replaced by https://github.com/Normation/ncf/pull/403

PR #403, opened 2016-07-11
Fixes #8660: Adapt documetation url of each method to us rudder documentation if editor is used with Rudder

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8660

Replacing previous PR: https://github.com/Normation/ncf/pull/402

### DIFF
--- a/api/ncf_api_flask_app/views.py
+++ b/api/ncf_api_flask_app/views.py
@@ -12,6 +12,7 @@ NCF_TOOLS_DIR = '/usr/share/ncf/tools'
 sys.path[0:0] = [NCF_TOOLS_DIR]
 
 default_path = ""
+use_rudder_auth = True
 
 import ncf
 import ncf_constraints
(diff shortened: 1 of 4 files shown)


VinceMacBuche:
Commit modified

api/ncf_api_flask_app/views.py
```diff
@@ -12,6 +12,7 @@
 sys.path[0:0] = [NCF_TOOLS_DIR]
 
 default_path = ""
+use_rudder_auth = True
```
fanf: Not sure about that value. Why Rudder by default in ncf? 
On the other hand, the auth was already set to Rudder by default...

VinceMacBuche: We already set Rudder as default auth method, so it easier to change it no (juste chnage the flag) 

Best way should be to have this value in a config file, but that would imply changing in the packaging too

VinceMacBuche:
Commit modified

PR #408, opened 2016-07-15
Fixes #8465: Limit length of technique names to 100 chars in the ncf api

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8465

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -410,8 +410,10 @@ def check_technique_method_call(bundle_name, method_call):
 
 
 def check_technique_metadata(technique_metadata):
-  """Check technique metdata, if one is missing raise an exception"""
+  """Check technique metdata, if one is missing raise an exception, returns an array of warnings"""
   mandatory_keys = [ 'name', 'bundle_name', 'method_calls' ]
+  warnings = []
+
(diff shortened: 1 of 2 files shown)


VinceMacBuche:
@fanf I ned to test it deeper with ncf-builder but code is almost ok

tools/ncf.py
```diff
@@ -421,13 +423,20 @@ def check_technique_metadata(technique_metadata):
     if technique_metadata[key] == "":
       raise NcfError("A technique must have a "+key+", but there is none")
 
+  # File name of the technique will be invalid on most filesystem, do not save technique
+  if len(technique_metadata['bundle_name']) > 253:
+    raise NcfError("Bundle names longer than 255 characters won't work on most filesystems.")
+  elif len(technique['bundle_name']) > 100:
+      warnings.push("Bundle names longer than 100 characters may not work on some filesystems (Windows, in particular).")
```
ncharles: what happenswhen this warning appears ? Does the user need to confirm something ?
If not, technique is written, but is there a way for the user to change the lengrh of the technique filename afterward ?

VinceMacBuche:
Commit modified

PR #411, opened 2016-07-20
Fixes #8721:  Broken service_ensure_running_path test with CFEngine 3.6.7, 3.7.3

amousset:
https://www.rudder-project.org/redmine/issues/8721


amousset:
Commit modified

PR #412, opened 2016-07-26
Fixes #8754: file ensure keys-> value cannot handle keys with special characters (like *)

ncharles:
https://www.rudder-project.org/redmine/issues/8754

### DIFF
--- a/tree/30_generic_methods/file_ensure_key_value.cf
+++ b/tree/30_generic_methods/file_ensure_key_value.cf
@@ -39,7 +39,7 @@ bundle agent file_ensure_key_value(file, key, value, separator)
   files:
       "${file}"
         create        => "true",
-        edit_line     => maintain_key_values("file_ensure_key_value.keys", "${separator}"),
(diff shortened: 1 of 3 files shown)


tree/30_generic_methods/file_ensure_key_value.cf
```diff
@@ -39,7 +39,7 @@ bundle agent file_ensure_key_value(file, key, value, separator)
   files:
       "${file}"
         create        => "true",
-        edit_line     => maintain_key_values("file_ensure_key_value.keys", "${separator}"),
+        edit_line     => ncf_maintain_keys_values("file_ensure_key_value.keys", "${separator}"),
```
peckpeck: Why do you replace key_values with keyS_values ?

ncharles: because it can set several keys

ncharles: see tree/30_generic_methods/file_ensure_keys_values.cf

jooooooon:
Ping @peckpeck ? It would be good to get this in the release.

PR #416, opened 2016-08-09
Fixes #8826: Technique Editor is broken in Rudder 3.2.5

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8826


VinceMacBuche:
Commit modified

tools/ncf_rudder.py
```diff
@@ -324,7 +324,7 @@ def generate_rudder_reporting(technique):
     key_value_canonified = regex.sub("_", key_value)
 
     class_prefix = generic_method["class_prefix"]+"_"+key_value_canonified
-    logger_rudder_call = '"dummy_report" usebundle => log_rudder("' + generic_method['name'] + ' ' + key_value + ' if ' + method_call['class_context'] + '", "' + class_prefix +'")'
+    logger_rudder_call = '"dummy_report" usebundle => log_rudder("' + generic_method['name'] + ' ' + key_value + ' if ' + method_call['class_context'] + '", "' + class_prefix +'", "", "")'
```
jooooooon: The last two arguments of log_rudder are actually important. I'm not sure how well the method would cope with the last (which is supposed to be a slist) being an empty string. Therefore, I think now is a good time to get ncf-builder to include the boilerplate to generate them in all methods, just like we do in the methods ncf provides.

That simply involves adding this code to the top of all generated bundles (before "methods:") (like in ncf.py):

```
  vars:
    "promisers"          slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
    "class_prefix"      string => canonify(join("_", "promisers"));
    "args"               slist => { };
```

(obviously, args will be empty for now since ncf-builder doesn't support creating techniques with arguments, but would later contain the list of arguments)

Then, this call would instead become:

```
logger_rudder_call = '"dummy_report" usebundle => log_rudder("' + generic_method['name'] + ' ' + key_value + ' if ' + method_call['class_context'] + '", "' + class_prefix +'", "${class_prefix}", @{args})'

```

VinceMacBuche:
Commit modified

jooooooon:
Looks good to me, but you mentioned it may not work. Did you test?

VinceMacBuche:
I tested and it works so ...

jooooooon:
Great!

PR #418, opened 2016-08-09
Fixes #8829: New method calls are considered valid, although they are empty which leads to error when validated via API

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8829


jooooooon:
Well... I don't understand all of this but it looks OK to me. I'll let you decide if you need a second opinion, @VinceMacBuche. Please merge if you're OK with this.

VinceMacBuche:
Maybe it's better for @fanf feedback next week!

VinceMacBuche:
OK, merging this PR

PR #424, opened 2016-09-01
Fixes #8896: We need to save twice after cloning a technique

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8896


VinceMacBuche:
Commit modified

PR #427, opened 2016-09-06
Fixes #8952: Add generic methods using new package promises

amousset:
https://www.rudder-project.org/redmine/issues/8952


amousset:
WIP

tree/30_generic_methods/package_present.cf
```diff
@@ -0,0 +1,111 @@
...
+#####################################################################################
+
+# @name Package present
+# @agent_version >=3.7
+# @description Enforce or verify the presence of a package
+# @documentation *Example*:
+# ```
+# methods:
+#     "any" usebundle => package_present("postgresql", "9.1", "x86_64", "enforce");
+#     "any" usebundle => package_present("postgresql", "lastest", "default", "enforce");
```
jooooooon: s/lastest/latest/

amousset:
Commit modified

amousset:
Commit modified

tree/30_generic_methods/package_present.cf
```diff
@@ -0,0 +1,119 @@
...
+      "warn"            expression => strcmp("warn", "${action}"),scope => "namespace";
+      "version_specified"      not => strcmp("any", "${version}");
+      "version_latest"         not => strcmp("latest", "${version}");
+      "architecture_specified" not => strcmp("default", "${architecture}");
+
+  packages:
+
+    debian.architecure_specified.version_specified::
+      "${name}"
+          policy => "present",
```
ncharles: could you align the => , please ?
Thank you !

amousset:
Commit modified

tree/30_generic_methods/package_present.cf
```diff
@@ -0,0 +1,146 @@
...
+      "reports" usebundle => _log("${action_description} ${name}${architecture_description}${version_description}", "${old_class_prefix}", "${class_prefix}", @{args});
+
+    !(debian|redhat)::
+      "force failure class" usebundle => _classes_failure("${class_prefix}");
+      "force failure class" usebundle => _classes_failure("${old_class_prefix}");
+      "report failure"      usebundle => _log("Package ${name} - systems other than Debian-, and Red Hat-like are not currently supported", "${old_class_prefix}", "${class_prefix}", @{args});
+
+}
+
+body package_module apt_get
```
ncharles: shouldn't this body be in the ncf_lib?
Plus it seem it doesn't do anything on redhat like system (nor suse)

amousset: It is still a WIP, I will add Red Hat support after. The body is there for convenience, it will be moved in a "3.7 ncf_lib" after.

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

tree/30_generic_methods/package_absent.cf
```diff
@@ -0,0 +1,145 @@
...
+# 
+# @class_prefix package_present
+# @class_parameter name
+
+bundle agent package_absent(name, version, architecture)
+{
+
+  vars:
+      "canonified_name"  string => canonify("${name}");
+      "old_class_prefix" string => "package_present_${canonified_name}";
```
ncharles: shouldn't it be package_absent ?

tree/30_generic_methods/package_absent.cf
```diff
@@ -0,0 +1,145 @@
...
+
+  methods:
+
+    (debian|redhat)::
+      "reports" usebundle => _log("Absence of package ${name}${architecture_description}${version_description}", "${old_class_prefix}", "${class_prefix}", @{args});
+
+    !(debian|redhat)::
+      "force failure class" usebundle => _classes_failure("${class_prefix}");
+      "force failure class" usebundle => _classes_failure("${old_class_prefix}");
+      "report failure"      usebundle => _log("Absence of package ${name} - systems other than Debian-, and Red Hat-like are not currently supported", "${old_class_prefix}", "${class_prefix}", @{args});
```
ncharles: given that we have quite a few sles user, we should support zypper also, what do you think?

tree/modules/packages/apt_get
```diff
@@ -0,0 +1,361 @@
+#!/usr/bin/python
+
```
ncharles: how will it be distributed to the nodes ? I don't see anything related to copy/deploiment of this

tree/30_generic_methods/package_absent.cf
amousset: I think we should but it is not yet supported by masterfiles, so we have to write our own script for zypper.

ncharles: yeah, i realised only later in the PR the complexity of the task :/

amousset:
Commit modified

amousset:
Commit modified

tree/20_cfe_basics/package_lib.cf
```diff
@@ -0,0 +1,38 @@
...
+    # Query package updates only every 24 hours.
+    query_updates_ifelapsed => "1440";
+    # Query installed packages every 2 hours.
+    query_installed_ifelapsed => "240";
+}
+
+body package_module yum
+{
+    # Query package updates only every 24 hours.
+    query_updates_ifelapsed => "1440";
```
amousset: @ncharles How do do you think we should make this configurable?

ncharles: i can make a dumb answer: add parameters in the body
a more clever answer would be to have logical default, and be able to define them globally (it doesn't make sense to have different values for different package on a given node). So it may be a generic method "configure package update frequency" that would define it through a bundle common something.
What do you think of it ?

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Tests are incomplete, we need to improve the test module to check versions/architectures.

tree/30_generic_methods/package_present.cf
```diff
@@ -0,0 +1,118 @@
...
+# * *latest* which will ensure, at each run, that the package is at the latest available version.
+# 
+# The last parameter is the provider, which is documented in the next section.
+#
+# #### Package providers
+# 
+# This method supports several package managers. You can specify the package manager
+# you want to use or let the method choose the default for the local system.
+# 
+# The package providers include a caching systems for package information.
```
jooooooon: system*

tree/30_generic_methods/_package_state.cf
```diff
@@ -0,0 +1,229 @@
...
+# @documentation See package_present and package_absent for documentation.
+# @parameter name Name of the package, or path to a local package if state is present
+# @parameter version Version of the package, can be "latest" for latest version or "any" for any version (defaults to "any")
+# @parameter architecture Architecture of the package, can be an architecture name  or "default" (defaults to "default")
+# @parameter provider Package provider to use, can be "yum", "apt", "pkg" or "default" for system default package manager (defaults to "default")
+# 
+# @class_prefix package_${state}
+# @class_parameter name
+
+bundle agent package_state(name, version, architecture, provider, state)
```
jooooooon: This bundle name doesn't match the file name - there should be a leading _

tree/30_generic_methods/_package_state.cf
```diff
@@ -0,0 +1,229 @@
...
+      "class_prefix"     string => canonify(join("_", "promisers"));
+      "args"             slist  => { "${name}", "${version}", "${architecture}", "${provider}", "${state}" };
+
+    # Build string vars used for reporting
+
+    # State
+    state_present::
+      "state_descritpion" string => "Presence";
+    !state_present::
+      "state_descritpion" string => "Absence";
```
jooooooon: Typo in both of these definition: description*

jooooooon:
Please provide an update on the status of this.

amousset:
Current state is:
* Working package management for yum/rpm, apt/deb and FreeBSD's pkg (untested)
* Working test to check that we call the module with the right parameters. Note that we do not test the modules, as the one we use are from masterfiles. We may need to add support for this when we have custom modules, but that would also require to be able to launch ncf tests on various currently unsupported platforms. 
* Support for modules in Rudder update process is being done in https://github.com/Normation/rudder-techniques/pull/1029

What is missing:
* Zypper support, as it not available yet in CFEngine. We could use the yum script as basis as it allows using the same RPM commands.
* The test is still a bit incomplete, and we need to check the version and architecture too.

amousset:
Commit modified

amousset:
package

jooooooon:
It's time to implement drop-down / multiple selection lists in ncf builder :) cc @VinceMacBuche @RaphaelGauthier

jooooooon:
OK, thanks, Alexis. I'll ask @aanriot to test then.

amousset:
Commit modified

amousset:
Commit modified

amousset:
* Fixed a typo that broke specified architectures
* Improved the test.

amousset:
We still have two open questions:

## Distributing the modules

https://github.com/Normation/rudder-techniques/pull/1029.

## Configuring the modules

Each module needs a body like:
```
body package_module apt_get
{
    # Query list of installed packages from package manager
    query_installed_ifelapsed => "60";
    # Query list of available updates from package manager
    query_updates_ifelapsed => "1440";
}
```

We should probably have default values and can make them overridable:
* With a specific call/generic method
* Using `ncf.conf` 

I think I prefer the second, as it is really a global configuration parameter and not a part of any policy (like execution frequency for example).
Ideally, It would be great to have a setting in Rudder to set this parameter by node/group/global.

amousset:
Commit modified

amousset:
* Added the body for the `test` module to `package_lib.cf` as it is now in the `_package_state` bundle.

amousset:
Commit modified

amousset:
Added a zypper module based on the yum module. Commands and options are adapted, but the logic from parsing zypper output is not there yet.

amousset:
Commit modified

amousset:
Added parsing for the list of available updates. We still need:
* Build package arguments (probably not very different, still not tested)
* Handle installation from repositories, a bit more test needed for update/install handling

amousset:
Commit modified

tests/acceptance/30_generic_methods/package_present.cf
```diff
@@ -0,0 +1,57 @@
+bundle common acc_path
```
ncharles: You are missing the header of the file

tree/10_ncf_internals/modules/packages/apt_get
```diff
@@ -0,0 +1,361 @@
+#!/usr/bin/python
```
ncharles: shouldn't we mention somewhere that it comes from the CFEngine repository ?

tree/10_ncf_internals/modules/packages/yum
```diff
@@ -0,0 +1,408 @@
+#!/usr/bin/python
```
ncharles: same here

tree/10_ncf_internals/modules/packages/zypper
```diff
@@ -0,0 +1,394 @@
+#!/usr/bin/python
```
ncharles: shouldn't we add our header ?

tree/30_generic_methods/_package_state.cf
```diff
@@ -0,0 +1,260 @@
...
+      "version_specified"      not => strcmp("any", "${version}");
+
+      "state_present" expression => strcmp("present", "${state}");
+
+      # If architecture is not specified, do not add it to the promise.
+      # The package module will pick the default one for the local platform
+      "architecture_specified" not => strcmp("default", "${architecture}");
+
+      # Select the right package manager
+      # Always use the specified one, and select a default based on the OS
```
ncharles: i don't understand how the default one would be selected

amousset:
Commit modified

tree/30_generic_methods/_package_state.cf
amousset: because of the `"provider"     string => "${package_module_knowledge.platform_default}", if_match_regex => "default";` above.

amousset: But this comment should be moved.

tree/30_generic_methods/_package_state.cf
```diff
@@ -0,0 +1,260 @@
...
+      "version_description" string => "in any version";
+    version_specified.!version_latest::
+      "version_description" string => "in version ${version}";
+
+  defaults:
+      "version"      string => "any", if_match_regex => "";
+      "architecture" string => "default", if_match_regex => "";
+      "provider"     string => "default", if_match_regex => "";
+      # Select the default packager for this platform if "default"
+      "provider"     string => "${package_module_knowledge.platform_default}", if_match_regex => "default";
```
ncharles: can you chain the default (does it work as expected ?)

tree/30_generic_methods/_package_state.cf
ncharles: thanks

amousset:
Commit modified

amousset:
Commit modified

amousset:
OK, merging this PR

PR #428, opened 2016-09-07
Fixes #8762: The Category list is cut on small screen

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8762

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -119,7 +119,7 @@
         </div>
             
         <div  class="container Override row fill" ng-show="authenticated" >
-          <div class="col-md-2 ncf_col"> 
+          <div class="col-lg-2 col-md-2 col-sm-4 col-xs-12 ncf_col">
                 <div class="title">Library</div>
                 <h3 >Techniques</h3>
                 <table class="table table-hover table-condensed table-ncf" >
@@ -147,7 +147,7 @@
                 </table>
             </div>
 
-            <form class="form-horizontal editForm col-md-5 ncf_col" name="editForm" novalidate >
+            <form class="form-horizontal editForm col-lg-5 col-md-5 col-sm-8 col-xs-12 ncf_col" name="editForm" novalidate >
             <div class="jumbotron" ng-if="selectedTechnique === undefined" >
                 <h1>Technique editor</h1>
                 <p>Create a new technique or edit one  from the list on the left.</p>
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
This is a really very nice responsive chnage !! But it miss the initial goal of the issue !

It that you can't access to some  generic method categories in the add method screen when screen height is too small

VinceMacBuche:
It can be scrolled on small screen but still not on normal screen ;)

tools/ncf.py
```diff
@@ -23,7 +23,7 @@
 additional_path = ["/opt/rudder/bin","/usr/sbin","/usr/local"]
 
 # Verbose output
-VERBOSE = 0
+VERBOSE = 1
```
VinceMacBuche: Set back to non verbose mode please

tools/ncf.py
```diff
@@ -79,7 +79,7 @@ def check_output(command, env = {}):
     sys.stderr.write(error)
   else:
     if VERBOSE == 1:
-      sys.stderr.write("VERBOSE: Exception triggered, Command returned error code " + retcode + "\n")
+      sys.stderr.write("VERBOSE: Exception triggered, Command returned error code " + str(retcode) + "\n")
```
VinceMacBuche: Good catch !!

RaphaelGauthier:
PR updated

RaphaelGauthier:
PR updated

builder/index.html
```diff
@@ -394,16 +391,15 @@ <h4 class="panel-title">
                             </li>
                         </ul>
                     </div>
-                    <div class="col-md-3" class="bs-sidebar navbar" >
-                        <ul class="nav bs-sidenav" style="position:fixed; margin-top:68px;">
+                    <div class="col-lg-2 col-md-2 col-sm-2 col-xs-4 ncf_col" class="bs-sidebar navbar" ng-if="selectedMethod === undefined">
```
VinceMacBuche: I think you could have col-lg-1 col-md1 (and inscrease on to col-lg-4 on the methods) and reduce the padding on that column.

It's better to have more place on the left side

RaphaelGauthier: This is what happens when I set col-lg-1 to the "Categories" list :

col-lg-1

It works properly on very large screen ( width > ~1450px), but below this width, some categories names are larger than the column width :/

VinceMacBuche: You can gain like 20 / 25px 

It's not a problem if the lines are cut at the end

RaphaelGauthier: Oh, I didn't see... I'm going to try that !

RaphaelGauthier:
PR updated

builder/index.html
RaphaelGauthier: I remove the column padding, and reduce to 10px the categories names padding. That works pretty well, thanks !

RaphaelGauthier:
PR updated

RaphaelGauthier:
PR updated

PR #433, opened 2016-09-13
Fixes #8270: Missing reporting in "Service ensure running with service path" when the service is not running

amousset:
https://www.rudder-project.org/redmine/issues/8270

### DIFF
--- a/tree/30_generic_methods/service_ensure_running_path.cf
+++ b/tree/30_generic_methods/service_ensure_running_path.cf
@@ -57,7 +57,7 @@ bundle agent service_ensure_running_path(service_name, service_path)
         ifvarclass => "(real_path.service_check_running_${canonified_service_path}_kept)|(!real_path.service_check_running_${canonified_service_name})";
  
       "class copy if not running"
-        usebundle => _classes_copy("service_start_${canonified_service_path}", "${class_prefix}"),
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/service_ensure_running_path.cf
```diff
@@ -57,7 +57,7 @@ bundle agent service_ensure_running_path(service_name, service_path)
         ifvarclass => "(real_path.service_check_running_${canonified_service_path}_kept)|(!real_path.service_check_running_${canonified_service_name})";
  
       "class copy if not running"
-        usebundle => _classes_copy("service_start_${canonified_service_path}", "${class_prefix}"),
+        usebundle => _classes_copy("service_start_${canonified_service_name}", "${class_prefix}"),
```
peckpeck: Both could exist since depending on the real_path class, service_check_running_ps and service_check_running can be called.
So you should probably copy both.

amousset: I don't see cases where service_start would be called with the path, what do you refer to?

peckpeck: My bad, this is OK

amousset:
OK, merging this PR

PR #434, opened 2016-09-14
Fixes #8657: The test framework should be able to use list-compatible-input to only test compatible generic methods

amousset:
https://www.rudder-project.org/redmine/issues/8657


amousset:
WIP

PR #436, opened 2016-09-14
Fixes #8956: Specified action bodies override the dry-run policy

amousset:
https://www.rudder-project.org/redmine/issues/8956


amousset:
Commit modified

ncharles:
This is great, thank you
However, don't forget to create a similar change for rudder-technique, file rudder-lib.st

PR #440, opened 2016-09-15
Fixes #9070: Conditionnal loading does not work as expected

amousset:
https://www.rudder-project.org/redmine/issues/9070


peckpeck:
How could it work at all ?

amousset:
The `<` case did not work.

PR #442, opened 2016-09-16
Fixes #8972: Check/Verify package generic methods are actually installing packages

amousset:
https://www.rudder-project.org/redmine/issues/8972


ncharles:
This solution is much neater that what we used to have. Thank you

amousset:
OK, merging this PR

PR #444, opened 2016-09-21
Fixes #9094: Add a bundle to copy local modules into workdir/modules

amousset:
https://www.rudder-project.org/redmine/issues/9094


tree/10_ncf_internals/initialization.cf
```diff
@@ -0,0 +1,44 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# Initialization
+# This bundle will make necessary step to prepare ncf before running any configuration
+# It needs to be called just after loading ncf and its configuration.
+bundle agent init
```
amousset: Should we reuse the `init` bundle? Given its name, it seems it may have been placed there for this kind of thing, but we may want to just create a new bundle for this.

tree/10_ncf_internals/initialization.cf
```diff
@@ -0,0 +1,44 @@
...
+{
+
+  files:
+      "${sys.workdir}/modules/"
+        copy_from         => local_dcp("${me.framework_path}/10_ncf_internals/modules/"),
+        depth_search      => recurse("inf"),
+        perms             => m("700"),
```
amousset: Not using a generic method as all our `permissions_*` methods require user+group, which we want to keep as current user here, even if not root.

tree/10_ncf_internals/initialization.cf
```diff
@@ -0,0 +1,44 @@
...
+        action            => immediate,
+        move_obstructions => "true",
+        classes           => classes_generic("ncf_internals_modules_update"),
```
amousset: I think we need reports for this in Rudder. We may want to build an `ncf_init_*` class that aggregates all initialization steps to hide internal mechanisms, what is you opinion?

tree/10_ncf_internals/initialization.cf
```diff
@@ -0,0 +1,44 @@
...
+
+# Initialization
+# This bundle will make necessary step to prepare ncf before running any configuration
+# It needs to be called just after loading ncf and its configuration.
+bundle agent init
+{
+
+  files:
+      "${sys.workdir}/modules/"
+        copy_from         => local_dcp("${me.framework_path}/10_ncf_internals/modules/"),
```
peckpeck: me is not a good name and did not work before (we removed its usage)
isn't there a file control body or something similar that contain current file's path ?

amousset: Something like `${this.promise_filename}`? And then guess the relative path of the modules folder? I don't like have different ways to find ncf path in ncf, but instead substitute the needed content of ncf's `promises.cf` in Rudder. I agree `me` is a bad name, was it used in Rudder before?

tree/10_ncf_internals/initialization.cf
jooooooon: Huh. We should create one.

tree/10_ncf_internals/initialization.cf
jooooooon: Yes, I think this is perfect content for the init bundle.

tree/10_ncf_internals/initialization.cf
jooooooon: `me.framework` is defined as `${sys.workdir}/inputs` so that's no going to work as is.

There is also `${this.promise_dirname}` that could be good to get this. I propose that you make a relative path from there.

peckpeck: There is no different way in ncf at the moment since there is nothing using full path.
There is a single way in rudder that is hardcoded and external to ncf.

If you don't want to autodetect it, avoid having an external dependency and use a parameter to the bundle instead.

amousset:
Commit modified

tree/10_ncf_internals/initialization.cf
amousset: Makes sense, updated (BTW we currently use `me.framework` path in the deprecated `file_template_expand.cf`, we should find a way to remove it from Rudder).

amousset:
Commit modified

PR #445, opened 2016-09-22
Fixes #8981: Cannot save new generic method with some contraints

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8981

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -89,6 +89,9 @@ app.directive('constraint', function($http, $q, $timeout) {
     },
     require: 'ngModel',
     link: function(scope, elm, attrs, ctrl) {
+    if((scope.parameter.constraints.allow_empty_string)&&(scope.parameter.value === undefined)){
+      scope.parameter.value = "";
+    }
     ctrl.$asyncValidators.constraint = function(modelValue, viewValue) {
       scope.parameter.$errors= [];
       if (modelValue === undefined) {


builder/js/ncf.js
```diff
@@ -89,6 +89,9 @@ app.directive('constraint', function($http, $q, $timeout) {
     },
     require: 'ngModel',
     link: function(scope, elm, attrs, ctrl) {
+    if(scope.parameter.constraints.allow_empty_string){
```
VinceMacBuche: You need to check if the value is undefined before unless value will always be empty

RaphaelGauthier:
PR updated

builder/js/ncf.js
RaphaelGauthier: Done!

builder/js/ncf.js
```diff
@@ -89,6 +89,9 @@ app.directive('constraint', function($http, $q, $timeout) {
     },
     require: 'ngModel',
     link: function(scope, elm, attrs, ctrl) {
+    if((scope.parameter.constraints.allow_empty_string)&&(scope.parameter.value==null)){
```
VinceMacBuche: should check for null but ( scope.parameter.value === undefined)

And ndever double equal ;)

RaphaelGauthier: I thought it was better to check both cases, null and undefined. But ok, it's done!

RaphaelGauthier:
PR updated

PR #446, opened 2016-09-22
Fixes #9105: Add conditions on features in ncf code loading

peckpeck:
https://www.rudder-project.org/redmine/issues/9105

### DIFF
--- a/tree/10_ncf_internals/list-compatible-inputs
+++ b/tree/10_ncf_internals/list-compatible-inputs
@@ -5,24 +5,58 @@ set -e
 # Need at least 3 parameters
 if [ "$#" -lt 3 ]
 then
-  echo "Usage $0 <CFEngine_version> <ncf_framework_path> [common|local]/<nn_directory> [<nn_directory ...]"
-  echo "  Finds all *.cf files in <ncf_framework>/[common|local]<nn_directory> that are compatible with the given CFEngine version"
+  echo "Usage: $0 [--capability-file <capability_file_path>] [--agent-version <CFEngine_version>] --ncf-path <ncf_framework_path> [common|local]/<nn_directory> [<nn_directory ...]"
+  echo "Deprecated usage: $0 <CFEngine_version> <ncf_framework_path> [common|local]/<nn_directory> [<nn_directory ...]"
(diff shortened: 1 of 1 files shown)


peckpeck:
Commit modified

peckpeck:
Commit modified

peckpeck:
Commit modified

peckpeck:
Commit modified

tree/10_ncf_internals/list-compatible-inputs
```diff
@@ -5,25 +5,56 @@ set -e
...
+else
+  cfengine_version="$1"
+  framework_path="$2"
+  # see comment below on the last parameters
+  shift 2
+fi
+
 # split version numbers
 version_regex='\([0-9][0-9]*\)\.\([0-9][0-9]*\).*'
 cfengine_major=`printf "${cfengine_version}\n" | sed -e "s/${version_regex}/\\1/"`
```
amousset: Shouldn't we check cfengine_version existence?

tree/10_ncf_internals/list-compatible-inputs
```diff
@@ -62,31 +93,68 @@ do
...
+            include_reqs=1 # false in shell
+          fi
+        done
+      else
+        # no capabilities required = accept
+        include_reqs=0 # true for a shell
+      fi
+
+      # merge and exclude if necessary
+      if [ ${include_version} -ne 0 ] || [ ${include_reqs} -ne 0 ]
```
amousset: If we have `agent_version >= 3.9` and `agent_requirements  capabilities: [ dumpdatastate ]`, the file would not be included on 4.0. I think we need an && here, as we are matching negative conditions.

tree/10_ncf_internals/list-compatible-inputs
peckpeck: yes we should

tree/10_ncf_internals/list-compatible-inputs
peckpeck: Be careful, this is a negative expression, and it uses shell convention for true/false.

It reads: if we exclude for version or if we exclude for requirements, then exclude ...

peckpeck:
Commit modified

amousset:
I think we need to rename the cache file to add the capabilities.

peckpeck:
Commit modified

PR #457, opened 2016-09-28
Fixes #9132: Package check installed does not produce reports

amousset:
https://www.rudder-project.org/redmine/issues/9132


amousset:
OK, merging this PR

PR #458, opened 2016-09-28
Fixes #9164: Uneeded scrollbar when using ncf builder in an iframe

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9164


RaphaelGauthier:
PR updated

PR #460, opened 2016-09-29
Fixes #9187: logger_rudder evaluates at each call the expected_reports_temp file

ncharles:
https://www.rudder-project.org/redmine/issues/9187

### DIFF
--- a/tree/30_generic_methods/logger_rudder.cf
+++ b/tree/30_generic_methods/logger_rudder.cf
@@ -27,6 +27,21 @@
 
 bundle agent logger_rudder(message, class_prefix)
 {
+  # The different files used for reporting are:
+  # * expected_reports_source: part of the promises, generated on the server
+  # * expected_reports_temp: temporary file templated from expected_reports_source
+  # * expected_reports_file: final file, generated from the content of expected_reports_temp
+  #
+  # The general workflow during the first call to logger_rudder of each run is:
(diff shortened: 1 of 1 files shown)


ncharles:
still testing, do not merge

ncharles:
tested, looks ok - on a small test system with standart directive + 3 ncf techniques with a total of 19 components, run length went from 11s to 9s

amousset:
Could we add something like this at the beginning of the bundle?

```
  # The different files used for reporting are:
  # * expected_reports_source: part of the promises, generated on the server
  # * expected_reports_temp: temporary file templated from expected_reports_source
  # * expected_reports_file: final file, generated from the content of expected_reports_temp
  #
  # The general workflow during the first call to logger_rudder of each run is:
  # 1/ Read the report file present in the promises (expected_reports_source),
  #    and expand it into the temporary report file (expected_reports_temp)
  # 2/ Read the content of the temporary file into a variable (reports)
  # 3/ Using the content of the read variable, write the content of the final file
  # 4/ Read the content of the final file into another variable (report_data), and use it for actual reporting
  #
  # The workflow for the other calls is only:
  #  4/ Read the content of the final file into another variable (report_data), and use it for actual reporting
```

ncharles:
Commit modified

ncharles:
it's done

ncharles:
OK, merging this PR

PR #463, opened 2016-09-30
Fixes #9199: Add unit test for rudder_logger

ncharles:
https://www.rudder-project.org/redmine/issues/9199


ncharles:
Commit modified

PR #466, opened 2016-10-05
Fixes #9226: Define a standard abort class in ncf

amousset:
https://www.rudder-project.org/redmine/issues/9226

### DIFF
--- a/tree/promises.cf
+++ b/tree/promises.cf
@@ -76,6 +76,9 @@ body agent control {
     # Repository where to put the copy of modified files
     default_repository => "${sys.workdir}/modified-files";
 
+    # Standard abort class usable in ncf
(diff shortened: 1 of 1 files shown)


tree/promises.cf
```diff
@@ -76,6 +76,9 @@ body agent control {
     # Repository where to put the copy of modified files
     default_repository => "${sys.workdir}/modified-files";
 
+    # Standard abort class usable in ncf
+    # The agent will stop execution with a fatal error as soon as it is defined
+    abortclasses => { "abort_agent_run" };
```
peckpeck: Why not use the same one as rudder ?

amousset: Because it is named `should_not_continue` which is a pretty bad name for something that stops the agent immediately with a fatal error.

PR #467, opened 2016-10-05
Fixes #9230: Define a dry-run lib when default bodies are not supported

amousset:
https://www.rudder-project.org/redmine/issues/9230

### DIFF
--- a/tree/20_cfe_basics/dry_run_lib.cf
+++ b/tree/20_cfe_basics/dry_run_lib.cf
@@ -16,13 +16,33 @@
 #
 #####################################################################################
 
-# There is 3 ways to have a promise dry-runned:
-#  - set the global_dry_run class (this one is not set by ncf)
-#  - call set_dry_run_mode()
+# The dry-run lib is based on classes, so you need to control the environment of promises to
+# set the dry-run mode.
+# To enable the dry-run mode after some point in the execution flow, call set_dry_run_mode("true")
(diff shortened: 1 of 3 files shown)


amousset:
This requires changes in `list-compatible-inputs` before merge (http://www.rudder-project.org/redmine/issues/9225).

amousset:
Commit modified

tree/20_cfe_basics/dry_run_lib.cf
```diff
@@ -34,6 +54,11 @@
 # WARNING: * promise type specific dry-run is NOT considered by action bodies, and will be overriden
 # WARNING:   when using any action body.
 # WARNING:
+#
+# @parameter mode   The dry-run mode, "true" to enable dry-run-mode and "false" to disable it
+# @agent_requirements "capabilities": [ "defaultbody" ] || "agent_version"" >= 3.9
```
peckpeck: This kind of expression is not yet supported

amousset: I know, but we will need it for this. We need to choose the syntax and implement it in list-compatible-inputs.

peckpeck: It was not implemented because it is very long and complex to implement in standard shell, I hope we can find time to implement it before the end of times

amousset:
Commit modified

PR #479, opened 2016-10-18
Fixes #9431: http* methods should allow enmpty headers field

amousset:
https://www.rudder-project.org/redmine/issues/9431


amousset:
Commit modified

PR #484, opened 2016-11-02
Fixes #9592: Improve jinja2 templating documentation

amousset:
https://www.rudder-project.org/redmine/issues/9592


ncharles:
thank you so much

PR #497, opened 2016-12-01
Fixes #9547: package_install should use another package body when using a specific package version

amousset:
https://www.rudder-project.org/redmine/issues/9547

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -249,6 +249,176 @@ archlinux::
  package_update_command => "/usr/bin/pacman -S --noconfirm --noprogressbar --needed";
 }
 
+# Same as ncf_generic with version in package name
+body package_method ncf_generic_version
+{
+  SuSE::
+    package_changes => "bulk";
+    package_list_command => "/bin/rpm -qa --queryformat \"i | repos | %{name} | %{version}-%{release} | %{arch}\n\"";
+    # set it to "0" to avoid caching of list during upgrade
+    package_list_update_command => "/usr/bin/zypper refresh";
+    package_list_update_ifelapsed => "240"; # 4 hours
+    package_installed_regex => "i.*";
+    package_list_name_regex    => "[^|]+\|[^|]+\|\s+([^\s]+).*";
+    package_list_version_regex => "[^|]+\|[^|]+\|[^|]+\|\s+([^\s]+).*";
+    package_list_arch_regex    => "[^|]+\|[^|]+\|[^|]+\|[^|]+\|\s+([^\s]+).*";
+    package_patch_name_regex    => "[^|]+\|\s+([^\s]+).*";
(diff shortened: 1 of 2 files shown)


amousset:
Do not merge for now, only SLES version has bee tested.

amousset:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -249,6 +249,176 @@ archlinux::
...
+
+  freebsd::
+    package_changes => "individual";
+    package_list_command => "/usr/sbin/pkg_info";
+    package_list_name_regex    => "([^\s]+)-.*";
+    package_list_version_regex => "[^\s]+-([^\s]+).*";
+    package_name_regex    => "([^\s]+)-.*";
+    package_version_regex => "[^\s]+-([^\s]+).*";
+    package_installed_regex => ".*";
+    package_name_convention => "$(name)-$(version)";
```
ncharles: this was already correct :)

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -249,6 +249,176 @@ archlinux::
...
+    package_add_command => "/sbin/apk add";
+    package_delete_command => "/sbin/apk del";
+
+  gentoo::
+    package_changes => "individual";
+    package_list_command => "/bin/sh -c '/bin/ls -d /var/db/pkg/*/* | cut -c 13-'";
+    package_list_name_regex => ".*/([^\s]+)-\d.*";
+    package_list_version_regex => ".*/[^\s]+-(\d.*)";
+    package_installed_regex => ".*";                          # all reported are installed
+    package_name_convention => "$(name)";
```
ncharles: aren't there version on gentoo ?

tree/30_generic_methods/package_install_version_cmp_update.cf
```diff
@@ -141,7 +141,7 @@ bundle agent package_install_version_cmp_update(package_name, version_comparator
                         package_policy  => "${update_policy}",
                         package_version => "${package_version}",
                         package_select  => "${version_comparator}",
-                        package_method  => ncf_generic,
+                        package_method  => ncf_generic_version,
```
ncharles: Could you use ncf_generic_version for previous case also, so that we have a consistent implementation

tree/20_cfe_basics/ncf_lib.cf
amousset: yes, and in the technique too (but I guess this is because the dash does not break anything when $(version) is empty, contrary to the equal sign...)

amousset:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
ncharles: no, it is not
it was a genuine question

amousset:
OK, merging this PR

PR #498, opened 2016-12-02
Fixes #9757: Broken log_rudder with CFEngine 3.10.0b1

amousset:
https://www.rudder-project.org/redmine/issues/9757


amousset:
Commit modified

PR #499, opened 2016-12-05
Fixes #9768: Modules are not correctly copied on Windows

ncharles:
https://www.rudder-project.org/redmine/issues/9768


ncharles:
OK, merging this PR

PR #500, opened 2016-12-05
Fixes #9772: promises.cf should use the new initialization bundle

amousset:
https://www.rudder-project.org/redmine/issues/9772


amousset:
Commit modified

amousset:
Commit modified

PR #501, opened 2016-12-06
Fixes #9779: Save button is not disable during saving 

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9779

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -194,8 +194,6 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location, 
 
   $scope.CForm = {};
 
-  $scope.saving = false;
-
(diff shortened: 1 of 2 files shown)


builder/js/ncf.js
```diff
@@ -847,23 +849,22 @@ $scope.groupMethodsByCategory = function () {
         $scope.techniques[index] = savedTechnique;
       }
     }
-    var saveError = function(action, data) {
-      $scope.saving = false;
+    $scope.saveError = function(action, data) {
```
VinceMacBuche: Can you switch back to a var here ? or a function directly !! 

would be simpler !!

RaphaelGauthier:
Commit modified

PR #503, opened 2016-12-15
Fixes #9806: Broken apt_get package module on Ubuntu 16.04

amousset:
https://www.rudder-project.org/redmine/issues/9806


amousset:
Commit modified

PR #520, opened 2017-01-26
Fixes #10079: Add a generic method to merge two dict variables

amousset:
https://www.rudder-project.org/redmine/issues/10079

### DIFF
--- a/tree/30_generic_methods/variable_dict_merge.cf
+++ b/tree/30_generic_methods/variable_dict_merge.cf
@@ -1,0 +1,97 @@
+#####################################################################################
+# Copyright 2017 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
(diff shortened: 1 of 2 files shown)


amousset:
Commit modified

peckpeck:
I thought this already existed

amousset:
Me too, but it seems not...

tree/30_generic_methods/variable_dict_merge.cf
```diff
@@ -0,0 +1,97 @@
...
+#   "key1": "different",
+#   "key2": "value2",
+#   "key3": "value3",
+#   "key4": "value4"
+# }
+# ```
+#
+# @parameter variable_prefix The prefix of the variable name
+# @parameter variable_name   The variable to define, the full name will be variable_prefix.variable_name
+# @parameter first_variable  The first variable, that will be overriden if necessary (in the form variable_prefix.variable_name)
```
peckpeck: Could you use the word source instead of overridden to avoid confusion about the state of the variable after the call

amousset:
Commit modified

PR #527, opened 2017-02-09
Fixes #10153: Calling twice file_ensure_key_value_present_in_ini_section for same file and section doesn't do anything for the second call

ncharles:
https://www.rudder-project.org/redmine/issues/10153


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #531, opened 2017-02-14
Fixes #10205: File edition is limited to 1MB file, which is really small

ncharles:
https://www.rudder-project.org/redmine/issues/10205


amousset:
Shouldn't we do it in Rudder too?

PR #469, opened 2016-10-07
Fixes #9236: Log levels should be incremental

amousset:
https://www.rudder-project.org/redmine/issues/9236

PR #532, opened 2017-02-15
Fixes #9236: Log levels should be incremental

amousset:
https://www.rudder-project.org/redmine/issues/9236

Replacing previous PR: https://github.com/Normation/ncf/pull/469

### DIFF
--- a/tree/10_ncf_internals/configuration.cf
+++ b/tree/10_ncf_internals/configuration.cf
@@ -79,6 +79,21 @@ bundle agent configuration
       "pass2" expression => "pass1";
       "pass1" expression => "any";
 
+      # Define ncf verbosity classes according to classes
+      # defined by the agent after verbosity options (-I, -v, -d)
+      "info"         expression => "inform_mode",
+                          scope => "namespace";
+      "debug"        expression => "verbose_mode",
+                          scope => "namespace";
+      "trace"        expression => "debug_mode",
+                          scope => "namespace";
+
+      # Make ncf log level incremental
+      "debug"        expression => "trace",
+                          scope => "namespace";
+      "info"         expression => "debug",
+                          scope => "namespace";
+
     pass3::
       # this line will not be evaluated by cfengine if destination_prefix contains a variable that does not exist
       # and it will always be evaluated to true otherwise
--- a/site/content/Quickstart.md
+++ b/site/content/Quickstart.md
@@ -71,6 +71,11 @@ Test it, run:
 
 Yep, it works !
 
+You can use the following classes to control the verbosity level:
+
+* `info`: for general information about the execution
+* `debug`: for detailed output (mainly to debug ncf itself)
+* `trace`: for very detailed output (mainly to debug ncf itself)
 
 Use ncf on a policy server:
 ---------------------------
--- a/tree/10_ncf_internals/dispatcher.cf
+++ b/tree/10_ncf_internals/dispatcher.cf
@@ -50,12 +50,13 @@ bundle agent dispatcher
   reports:
     !mapper_services_exists::
       "${configuration.error} There are no services defined in service_mapping.cf. The execution is stopped";
-    mapper_services_exists::
+    info.mapper_services_exists::
       "${configuration.info} Running services, going to run bundle: ${services_array_${services}[2]} (in version '${services_array_${services}[3]}')";
 
-    cfengine::
+    info::
       "${configuration.info} Starting bundle runs..."
         classes => classes_generic("dispatcher_reports");
+    any::
       "${configuration.error} Could not run bundle ${services_array_${services}[2]} (with service path ${services_array_${services}[0]})"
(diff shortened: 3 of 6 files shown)


amousset:
Commit modified

tree/10_ncf_internals/configuration.cf
```diff
@@ -79,6 +79,18 @@ bundle agent configuration
       "pass2" expression => "pass1";
       "pass1" expression => "any";
 
+      # Make log level incremental
+      "debug"        expression => "trace",
+                          scope => "namespace";
+      # CFEngine's stdlib verbose mode
+      "verbose_mode" expression => "debug",
+                          scope => "namespace";
+      "info"         expression => "debug",
```
peckpeck: shouldn' this be verbose not debug ?

peckpeck:
everything seems reverted, debug implies verbose etc ...

peckpeck:
Did you see this with jon ?

tree/10_ncf_internals/configuration.cf
amousset: the verbose class does not exist

amousset:
Commit modified

amousset:
Changed some reports to make things consistent: info level is only displayed when the info class is defined. This is closer to CFEngine behavior.

amousset:
Commit modified

amousset:
Also added some doc (in the quickstart as we do not have anything better for now).

tree/10_ncf_internals/configuration.cf
```diff
@@ -79,6 +79,18 @@ bundle agent configuration
       "pass2" expression => "pass1";
       "pass1" expression => "any";
 
+      # Make log level incremental
+      "debug"        expression => "trace",
```
peckpeck: I still see trace implies debug

tree/10_ncf_internals/configuration.cf
```diff
@@ -79,6 +79,18 @@ bundle agent configuration
...
+                          scope => "namespace";
+      # CFEngine's stdlib verbose mode
+      "verbose_mode" expression => "debug",
```
peckpeck: debug implies verbose

tree/10_ncf_internals/configuration.cf
peckpeck: debug implies info

amousset: What do you think should it be instead?

ncharles: verbose_mode is defined by running agent with -v ; so it should imply info

amousset: Interesting!

amousset: In ncf, we have:
* *info*
* *debug*
* *trace*

CFEngine has:
* *inform_mode*
* *verbose_mode*
* *debug_mode*

We should at least define ncf classes when CFEngine classes are defined.

peckpeck: you're right i reversed the implication in my head

So
- we say that rudder loglevel decides cfengine stdlib loglevel
- we say that upper loglevel implies lower loglevel

Where does the trace loglevel come from ?

peckpeck: So we should set debug_mode too here no ?

peckpeck: I've just seen your upper comment.
Don't we need to split debug and trace too in rudder agent ?

amousset:
Commit modified

amousset:
Commit modified

tree/10_ncf_internals/configuration.cf
```diff
@@ -79,6 +79,19 @@ bundle agent configuration
       "pass2" expression => "pass1";
       "pass1" expression => "any";
 
+      "info"         expression => "inform_mode",
```
amousset: @peckpeck @ncharles Now:

* Options defined by CFEngine will define the matching ncf verbosity class (but verbose in CFEngine = debug in ncf, and debug in CFEngine = trace in ncf)
* ncf classes are incrementally defined

peckpeck: Thanks, this is much easier to understand

peckpeck: in this case it means our class definition in rudder agent is redundant ?

amousset: Indeed, I did not know this class was set by CFEngine itself.

amousset: We might want to keep it for now though, for compatibility with older ncf.

amousset:
Commit modified

tree/10_ncf_internals/configuration.cf
peckpeck: you're right

PR #537, opened 2017-02-22
Fixes #10232: Improve service_action methods

amousset:
https://www.rudder-project.org/redmine/issues/10232


amousset:
Commit modified

PR #546, opened 2017-03-02
Fixes #10336: Do not use dots in share-files id

amousset:
https://www.rudder-project.org/redmine/issues/10336


amousset:
Commit modified

amousset:
Commit modified

PR #547, opened 2017-03-06
Fixes #10332: Document the behavior of service methods

amousset:
https://www.rudder-project.org/redmine/issues/10332


ncharles:
This is so awesome ! Thank you so much Alexis

PR #549, opened 2017-03-08
Fixes #10370: cfengine_classes.json does not contain json but a js script

amousset:
https://www.rudder-project.org/redmine/issues/10370


amousset:
Maybe it is better to make it a real json, but loading it goes beyond my js skills :)

VinceMacBuche:
That's a real json, but since it only contains definition of a javascript object but yeahy renaming may not be that bad

PR #553, opened 2017-03-23
Fixes #10488: Do not use systemd to manage legacy init scripts

amousset:
https://www.rudder-project.org/redmine/issues/10488

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -807,7 +807,7 @@ bundle agent ncf_services(service, action)
     # These commands (sadly) are not completely static (as on debian 8 and upstart the outcome depends on the service)
 
     # systemd
-    pass1.!broken_systemctl.systemctl_utility_present::
+    pass1.!is_init_service.systemctl_utility_present::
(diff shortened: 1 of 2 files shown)


amousset:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -901,7 +901,7 @@ bundle agent ncf_services(service, action)
 
     # init.d
     # WARN: Some actions may not be supported by the init script
-    pass1.!is_boot_action.!systemctl_utility_present.!is_upstart_service.!svcadm_utility_present.!service_utility_present.!startsrc_utility_present.init_d_directory_present::
+    pass1.!is_boot_action.!systemctl_utility_present.!is_upstart_service.!svcadm_utility_present.!service_utility_present.!startsrc_utility_present.is_init_service::
```
ncharles: this condition is invalid: it may be a systemctl_utiliy_present, but with a init script
it should be (!systemctl_utility_present|is_init_service) also

ncharles:
Awesome

amousset:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
amousset: Good catch! Indeed.

PR #557, opened 2017-03-27
Fixes #10508: Add a generic method to remove directories

amousset:
https://www.rudder-project.org/redmine/issues/10508

### DIFF
--- a/tests/acceptance/30_generic_methods/directory_absent.cf
+++ b/tests/acceptance/30_generic_methods/directory_absent.cf
@@ -1,0 +1,129 @@
+#######################################################
+#
+# Test checking that directory_absent removes a directory
+#
+#######################################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "tmp"               string => getenv("TEMP", 1024);
+    "directory"         string => "${tmp}/test";
+    "directory_canon_0" string => canonify("${directory}/subdir0");
+    "directory_canon_1" string => canonify("${directory}/subdir1/");
+    "directory_canon_2" string => canonify("${directory}/subdir2");
+    "directory_canon_3" string => canonify("${directory}/subdir3");
+    "directory_canon_4" string => canonify("${directory}/subdir4");
+    "directory_canon_5" string => canonify("${directory}/subdir5");
+    "directory_canon_6" string => canonify("${directory}/file");
+    "directory_canon_7" string => canonify("${directory}/file");
+
+  files:
+    "${directory}/."
+      create => "true";
+
+    "${directory}/file"
+      create => "true";
+
+    #----
+
+    "${directory}/subdir0/."
+      create => "true";
+
+    #----
+
+    "${directory}/subdir1/."
+      create => "true";
+
+    "${directory}/subdir1/subfile1"
+      create => "true";
+
+    #----
+
+    "${directory}/subdir2/."
+      create => "true";
+
+    "${directory}/subdir2/subfile2"
+      create => "true";
+
+    "${directory}/subdir2/subsubdir2/."
+      create => "true";
+
+    "${directory}/subdir2/subsubdir2/subsubfile2"
+      create => "true";
+
+    #----
+
+    "${directory}/subdir3/."
+      create => "true";
+
+    "${directory}/subdir3/subfile3"
+      create => "true";
+
+    "${directory}/subdir3/subsubdir3/."
+      create => "true";
+
+    "${directory}/subdir3/subsubdir3/subsubfile3"
+      create => "true";
+
+}
+
+#######################################################
+
+bundle agent test
+{
+  methods:
+    "ph0" usebundle => directory_absent("${init.directory}/subdir0", "true");
+    "ph1" usebundle => directory_absent("${init.directory}/subdir1/", "true");
+    "ph2" usebundle => directory_absent("${init.directory}/subdir2", "true");
+    "ph3" usebundle => directory_absent("${init.directory}/subdir3", "false");
+    "ph4" usebundle => directory_absent("${init.directory}/subdir4", "true");
+    "ph5" usebundle => directory_absent("${init.directory}/subdir5", "false");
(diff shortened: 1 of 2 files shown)


tree/30_generic_methods/directory_absent.cf
```diff
@@ -0,0 +1,85 @@
...
+      "directory_content_size" int => length("directory_content");
+
+  defaults:
+      "recursive"           string => "false", if_match_regex => "";
+
+  classes:
+      "target_exists"   expression => fileexists("${target}/.");
+      "is_recursive"    expression => strcmp("true", "${recursive}");
+    pass2::
+      "is_empty"        expression => strcmp("0", "${directory_content_size}");
```
ncharles: Shouldn't this check be delayed AFTER the deletion ?

tree/30_generic_methods/directory_absent.cf
```diff
@@ -0,0 +1,85 @@
...
+        file_select   => all,
+        classes       => classes_generic("${class_prefix}"),
+        ifvarclass    => "!${class_prefix}_inside_error.(is_recursive|is_empty)";
+
+  methods:
+    pass3::
+      # If not empty, we need to trigger an error manually as the files promise will be kept
+      # Because of https://tracker.mender.io/browse/CFE-1570
+      "fail_if_not_empty"  usebundle  => _classes_failure("${class_prefix}"),
+                           ifvarclass => "target_exists.!is_recursive.!is_empty";
```
ncharles: Risk is the behaviour of the agent changes, and so we get a reports based on the expected "non logical" behaviour, rather than actual one.
Could you delete in the folder in pass1, and check in pass2 content of folder, to be future-proof?

amousset: The idea is to only create those classes when needed (I.E. not empty target and non-recursive deletion) and avoid calling the files promise because it would create kept classes.

ncharles: Yes, but what is agent stops to refuse deleting non empty folder ?

amousset: It would break. We could do the check after and use classes_sanitize to remove non-error classes.

amousset: Made a seconf commit with inverted check logic. This makes the code quite more readable!

amousset: Another solution is to just base reporting on check of the existence of the directory before/after file promises and ignore their outcome.

tree/30_generic_methods/directory_absent.cf
```diff
@@ -0,0 +1,79 @@
...
+      # If target was not empty, we need to trigger an error manually as the files promise will be kept
+      # Because of https://tracker.mender.io/browse/CFE-1570
+      "fail_if_not_empty"   usebundle  => _classes_failure("${class_prefix}"),
+                            ifvarclass => "target_exists_after";
+      "clean_if_not_empty"  usebundle  => _classes_sanitize("${class_prefix}"),
+                            ifvarclass => "target_exists_after";
+
+      # Use classes of the first promise when it failed
+      "copy_inside_error"   usebundle  => _classes_copy("${class_prefix}_inside", "${class_prefix}"),
+                            ifvarclass => "${class_prefix}_inside_error";
```
ncharles: Awesome !
One last question: would it detect that it is a file and not a folder ?

amousset: We need a test for this anyway :)

amousset: And a special case:

```
File '/home/amousset/Rudder/ncf/tests/acceptance/workdir/__30_generic_methods_directory_absent_cf/tmp/test/file/.' does not exist as promised
```

amousset:
Added a commit that allows returning errors for files.

tree/30_generic_methods/directory_absent.cf
```diff
@@ -0,0 +1,79 @@
...
+      "class_prefix"            string => canonify("directory_absent_${target}");
+
+  defaults:
+      "recursive"               string => "false", if_match_regex => "";
+
+  classes:
+      "is_recursive"        expression => strcmp("true", "${recursive}");
+
+    pass2::
+      "target_exists_after" expression => fileexists("${target}");
```
ncharles: shouldn't you use isdir to avoid detecting file that would have the same name as folder?
https://docs.cfengine.com/docs/3.7/reference-functions-isdir.html

amousset: Using `${target}/.` we ensure that a file won't be removed. We can use isdir, but that means we also need to have an existence check to make different between a file and an absent directory. Then we need to skip files promises and define classes manually depending on the case. Do you think it would be better?

ncharles: I was not thinking of skipping the deletion, but of detecting after the fact that folder is still there (or not) - and that it is not a file with same name that would report

tests/acceptance/30_generic_methods/directory_absent.cf
```diff
@@ -0,0 +1,129 @@
...
+    "ok_1"          expression => "!file_exists_1.directory_absent_${init.directory_canon_1}_repaired.!directory_absent_${init.directory_canon_1}_error";
+    "ok_2"          expression => "!file_exists_2.directory_absent_${init.directory_canon_2}_repaired.!directory_absent_${init.directory_canon_2}_error";
+    "ok_3"          expression => "file_exists_3.!directory_absent_${init.directory_canon_3}_repaired.directory_absent_${init.directory_canon_3}_error";
+    "ok_4"          expression => "!file_exists_4.directory_absent_${init.directory_canon_4}_kept.!directory_absent_${init.directory_canon_4}_error";
+    "ok_5"          expression => "!file_exists_5.directory_absent_${init.directory_canon_5}_kept.!directory_absent_${init.directory_canon_5}_error";
+    "ok_6"          expression => "file_exists_6.!directory_absent_${init.directory_canon_6}_kept.directory_absent_${init.directory_canon_6}_error";
+    "ok_7"          expression => "file_exists_7.!directory_absent_${init.directory_canon_7}_kept.directory_absent_${init.directory_canon_7}_error";
```
ncharles: Hum, don't we expect a kept for this one ? Folder does not exist, so there is nothing to delete

amousset: There is a file with the same name, I don't think we can ignore it. We may see it as:

```
# rmdir LICENSE 
rmdir: failed to remove 'LICENSE': Not a directory
```

ncharles: Ha, this is something that need to be discussed then.

ncharles:
Looks good to me! Great work

ncharles:
can you rebase ?

amousset:
PR rebased.

amousset:
Commit modified

PR #562, opened 2017-03-31
Fixes #9663: ncf Editor: Highlight condition box if a condition is active

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9663

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -188,6 +188,9 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location, 
   $scope.selectedMethod;
   // Are we authenticated on the interface
   $scope.authenticated = false;
+  // Open/Close by default the Conditions box
+  $scope.conditionIsOpen = false;
+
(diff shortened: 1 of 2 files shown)


VinceMacBuche:
It will always be open here, don't we want to open it if there is a condition set ?

RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

builder/js/ncf.js
```diff
@@ -188,6 +188,9 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location,
   $scope.selectedMethod;
   // Are we authenticated on the interface
   $scope.authenticated = false;
+  // Open/Close by default the Conditions box
+  $scope.conditionIsOpen;
```
VinceMacBuche: You should define "false" as default value, to prevent undefined

RaphaelGauthier:
Commit modified

PR #566, opened 2017-04-06
Fixes #9867: Directive Package management says "Presence of package rudder-agent in latest available version was correct" but it's not

amousset:
https://www.rudder-project.org/redmine/issues/9867


amousset:
Commit modified

amousset:
Commit modified

PR #581, opened 2017-04-27
Fixes #9079: When a ncf technnique has an ampersand in its name, rudder doesn't see it

peckpeck:
https://www.rudder-project.org/redmine/issues/9079

### DIFF
--- a/tools/ncf_rudder.py
+++ b/tools/ncf_rudder.py
@@ -159,6 +159,14 @@ def write_rudder_reporting_file(path,technique):
   file.close()
 
 
+def escape_entities( str ):
+    str = str.replace("&", "&amp;")
+    str = str.replace("<", "&lt;")
+    str = str.replace(">", "&gt;")
+    str = str.replace("\"", "&quot;")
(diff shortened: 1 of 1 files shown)


tools/ncf_rudder.py
```diff
@@ -159,15 +159,22 @@ def write_rudder_reporting_file(path,technique):
   file.close()
 
 
+def escape_entities( str ):
+    str = str.replace("&", "&amp;")
+    str = str.replace("<", "&lt;")
+    str = str.replace(">", "&gt;")
+    str = str.replace("\"", "&quot;")
```
amousset: don't we need single quote too?

peckpeck: Attributes in xml should be put within double quote and that's what we do.

amousset: It seems to be only a common convention (http://stackoverflow.com/questions/6800467/quotes-in-xml-single-or-double), but we indeed use only double quotes.

peckpeck:
Commit modified

PR #586, opened 2017-05-18
Fixes #10749: Broken package_state reporting

amousset:
https://www.rudder-project.org/redmine/issues/10749


amousset:
This should work with Rudder reporting if there is only one method_* call by package name. It is not a good general fix because it would cause double logging with other loggers.

amousset:
Commit modified

amousset:
Do not merge for now, still testing.

amousset:
Commit modified

amousset:
@peckpeck Ready for review.

amousset:
Commit modified

amousset:
Commit modified

PR #589, opened 2017-06-14
Fixes #10716: sharedfile_to_node broken if file is updated after shared

amousset:
https://www.rudder-project.org/redmine/issues/10716


amousset:
Commit modified

PR #594, opened 2017-06-15
Fixes #10919: Use a specific package body when no version is specified for yum/rpm

amousset:
https://www.rudder-project.org/redmine/issues/10919

### DIFF
--- a/tree/30_generic_methods/package_install_version_cmp_update.cf
+++ b/tree/30_generic_methods/package_install_version_cmp_update.cf
@@ -122,7 +122,7 @@ bundle agent package_install_version_cmp_update(package_name, version_comparator
     version_not_specified.!local_package.redhat::
       "${package_name}"
                         package_policy => "${update_policy}",
-                        package_method => yum_rpm,
+                        package_method => yum_rpm_no_version,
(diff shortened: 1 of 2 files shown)


amousset:
Commit modified

tree/30_generic_methods/package_install_version_cmp_update.cf
```diff
@@ -106,7 +106,7 @@ bundle agent package_install_version_cmp_update(package_name, version_comparator
                         package_policy  => "${update_policy}",
                         package_version => "${package_version}",
                         package_select  => "${version_comparator}",
-                        package_method  => yum_rpm,
+                        package_method  => yum_rpm_no_version,
```
ncharles: it's at the wrong place, it should be changed where version_not_specified

amousset:
Commit modified

tree/30_generic_methods/package_install_version_cmp_update.cf
amousset: indded, fixed.

amousset:
Tested on RHEL6.

amousset:
OK, merging this PR

PR #602, opened 2017-06-21
Fixes #10972: Fix small commit error

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10972


amousset:
It would be great to add a test for this, it is really dangerous.

PR #609, opened 2017-07-03
Fixes #11021: Cannot generate ps1 technique file (missing 50_techniques directory)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11021

### DIFF
--- a/tools/ncf_dsc.py
+++ b/tools/ncf_dsc.py
@@ -25,10 +25,18 @@ from pprint import pprint
 
 # MAIN FUNCTIONS called by command line parsing
 ###############################################
+dsc_ncf_path = "/var/rudder/configuration-repository/dsc/ncf"
+dsc_techniques_path = os.path.join(dsc_ncf_path,"50_techniques")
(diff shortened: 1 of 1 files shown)


tools/ncf_dsc.py
```diff
@@ -25,10 +25,18 @@
 
 # MAIN FUNCTIONS called by command line parsing
 ###############################################
+dsc_ncf_path = "/var/rudder/configuration-repository/dsc/ncf"
+dsc_techniques_path = os.path.join(dsc_ncf_path,"50_techniques")
+rudder_ncf_path = "/var/rudder/configuration-repository/ncf"
+
+def check_dsc_techniques_path():
```
ncharles: why not doing it in the packaging ?

VinceMacBuche: So it fix it even if it deleted after install

VinceMacBuche: We do likewise on rudder-side (metadata.xml)

VinceMacBuche: To check if conf-repo/techniques/ncf_techniques exists

PR #610, opened 2017-07-03
Fixes #11079: runlevel command is caled without path

amousset:
https://www.rudder-project.org/redmine/issues/11079


amousset:
OK, merging this PR

PR #614, opened 2017-07-07
Fixes #11113: Save button is not disable during saving

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11113

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -925,7 +925,6 @@ $scope.groupMethodsByCategory = function () {
     // Update selected technique if it's still the same technique
     // update technique from the tree
     var saveSuccess = function(data, status, headers, config) {
-      $scope.$broadcast('endSaving');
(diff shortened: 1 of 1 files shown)


builder/js/ncf.js
```diff
@@ -191,6 +191,7 @@ app.controller('ncf-builder', function ($scope, $modal, $http, $log, $location,
   // Open/Close by default the Conditions box
   $scope.conditionIsOpen = false;
 
+  $scope.saving = "ok";
```
VinceMacBuche: What is the usage of $scope.saving ?

VinceMacBuche: Why "ok" and not "false" ?

RaphaelGauthier: Oops, my bad. It should be _false_ !!

RaphaelGauthier:
Commit modified

RaphaelGauthier:
PR rebased

PR #618, opened 2017-07-26
Fixes #11174: Add new method to get file from Rudder shared folder

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11174


VinceMacBuche:
Commit modified

PR #621, opened 2017-08-04
Fixes #11211: Add a classic tag to generic methods

peckpeck:
https://www.rudder-project.org/redmine/issues/11211

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -577,6 +577,14 @@ def get_all_techniques_metadata(include_methods_calls = True, alt_path = ''):
   return { "data": { "techniques" : all_metadata }, "errors": format_errors(errors), "warnings": warnings }
 
 
+def get_agents_support(method, content):
+  agents = []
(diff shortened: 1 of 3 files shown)


tools/ncf.py
```diff
@@ -577,6 +577,14 @@ def get_all_techniques_metadata(include_methods_calls = True, alt_path = ''):
   return { "data": { "techniques" : all_metadata }, "errors": format_errors(errors), "warnings": warnings }
 
 
+def get_agents_support(method, content):
+  agents = []
+  if os.path.exists("/var/rudder/configuration-repository/dsc/ncf/30_generic_methods/" + method + ".ps1"):
+    agents.append("dsc")
+  if not re.search(r'\n\s*bundle\s+agent\s+'+method+r'\b.*?\{\s*\}', content, re.DOTALL): # this matches an empty bundle content
```
amousset: The `.cf` version of dsc-only methods seems not to be empty but contains a log (https://github.com/Normation/rudder-agent-windows/blob/master/plugin/ncf/30_generic_methods/service_status.cf for example)?

PR #624, opened 2017-08-10
Fixes #11245: Add constraint on hash_type

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11245

### DIFF
--- a/tree/30_generic_methods/file_from_shared_folder.cf
+++ b/tree/30_generic_methods/file_from_shared_folder.cf
@@ -21,7 +21,8 @@
 #
 # @parameter source      Source file (path relative to Rudder shared folder)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated
+# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256). Only used on dsc agent, cfengine agent use it's own system for now. 
+# @parameter_constraint hash_type  "select" : [ "md5", "sha1", "sha256" ]
 # 
 # @class_prefix file_from_shared_folder
 # @class_parameter destination


tree/30_generic_methods/file_from_shared_folder.cf
```diff
@@ -21,7 +21,8 @@
 #
 # @parameter source      Source file (path relative to Rudder shared folder)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated
+# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256)
```
peckpeck: You should document that this parameter is always treated as md5 on cfengine.

VinceMacBuche:
Commit modified

tree/30_generic_methods/file_from_shared_folder.cf
```diff
@@ -21,7 +21,8 @@
 #
 # @parameter source      Source file (path relative to Rudder shared folder)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated
+# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256), only md5 can be used on cfengine agent
```
ncharles: it's not used on cfengine, so why not simply state that it is not used for cfengine agent ?

VinceMacBuche: I will write that on cfengine it will always be treated as md5 for now

ncharles: ping ?

VinceMacBuche:
Commit modified

PR #629, opened 2017-08-21
Fixes #11215: Add "Classic" icon on generic methods compatible with the DSC and the Classic agent

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11215

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -296,7 +296,8 @@
                                     <a ng-click="$event.stopPropagation();" target="_blank" ng-href="http://www.ncf.io/pages/reference.html#{{getMethodBundleName(method_call)}}">
                                       <i class="glyphicon glyphicon-info-sign method-actions"></i>
                                     </a>
-                                    <i class="{{getIconClassByAgentType(method_call.method_name)}}"></i>
+                                    <i class="dsc-icon" ng-if="checkMethodCallAgentSupport(method_call.method_name,'dsc')"></i>
+                                    <i class="glyphicon glyphicon-cog classic-icon" ng-if="checkMethodCallAgentSupport(method_call.method_name,'cfengine-community')"></i>
                                   </span>
                                 <br style="clear:both"/>
                                 <small class="pull-left method-margin" >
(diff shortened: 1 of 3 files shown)


builder/index.html
```diff
@@ -540,8 +541,8 @@ <h4 class="panel-title">
                                     <i class="glyphicon glyphicon-info-sign method-actions"></i>
                                   </a>
                                 </span>
-                                <i class="dsc-icon" ng-if="method.dsc_support"></i>
-                                <i class="glyphicon glyphicon-cog classic-icon" ng-if="!method.dsc_support"></i>
+                                <i class="dsc-icon" ng-if="checkAgentSupport(method,'dsc')"></i>
+                                <i class="glyphicon glyphicon-cog classic-icon" ng-if="checkAgentSupport(method,'classic')"></i>
```
VinceMacBuche: change to cfengine-community instead of classic

builder/index.html
```diff
@@ -296,7 +296,8 @@ <h4 class="panel-title">
                                     <a ng-click="$event.stopPropagation();" target="_blank" ng-href="http://www.ncf.io/pages/reference.html#{{getMethodBundleName(method_call)}}">
                                       <i class="glyphicon glyphicon-info-sign method-actions"></i>
                                     </a>
-                                    <i class="{{getIconClassByAgentType(method_call.method_name)}}"></i>
+                                    <i class="dsc-icon" ng-if="checkMethodCallAgentSupport(method_call.method_name,'dsc')"></i>
+                                    <i class="glyphicon glyphicon-cog classic-icon" ng-if="checkMethodCallAgentSupport(method_call.method_name,'classic')"></i>
```
VinceMacBuche: You should use checkAgent instead

builder/js/ncf.js
```diff
@@ -591,22 +589,20 @@ $scope.groupMethodsByCategory = function () {
   $scope.checkFilterCategory = function(methods){
```
VinceMacBuche: That method should not be applied like this, but using  a filter on the needed ng-repeat https://docs.angularjs.org/api/ng/filter/filter

RaphaelGauthier: As the needed ng-repeat loop through an object, we can't, because Angular allow to use custom filter only on array, not on object.
> Selects a subset of items from **array** and returns it as a new array.

RaphaelGauthier:
Commit modified

builder/js/ncf.js
```diff
@@ -588,60 +585,50 @@ $scope.groupMethodsByCategory = function () {
   $scope.checkDeprecatedFilter = function(methods){
     return methods.some(function(method){return method.deprecated === undefined });
   }
+
   $scope.checkFilterCategory = function(methods){
```
VinceMacBuche: At least that should be applied on the method!

Nevermind, we will work this out in another PR

PR #631, opened 2017-08-22
Fixes #11283: automatically generate na report for method call that does not support cfengine-community in techniques

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11283

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -380,10 +380,11 @@ def parse_technique_methods(technique_file):
       if ifvarclass_context is not None:
         promise_class_context = class_context_and(class_context, ifvarclass_context)
 
-      if args:
-        res.append({'class_context': promise_class_context, 'promiser': promiser, 'method_name': method_name, 'args': args})
-      else:
-        res.append({'class_context': promise_class_context, 'promiser': promiser, 'method_name': method_name})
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

tools/ncf.py
```diff
@@ -380,10 +380,11 @@ def parse_technique_methods(technique_file):
       if ifvarclass_context is not None:
         promise_class_context = class_context_and(class_context, ifvarclass_context)
 
-      if args:
-        res.append({'class_context': promise_class_context, 'promiser': promiser, 'method_name': method_name, 'args': args})
-      else:
-        res.append({'class_context': promise_class_context, 'promiser': promiser, 'method_name': method_name})
+      if not (method_name.startswith("_") or method_name.startswith("log")):
```
amousset: why?

VinceMacBuche: To prevent treating _class and log methods in the technique which fails in most case when reading the metadata of these methods, i needed it when na reports were in the technique file

Maybe we can remove that line

PR #634, opened 2017-08-29
Fixes #11313: "Techniques diverge" popup appears after saving a technique if we added a new Generic methods

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11313


RaphaelGauthier:
Commit modified

PR #638, opened 2017-09-12
Fixes #11346: Create a new generic method to set key/value with option to define if it is strict or not

ncharles:
https://www.rudder-project.org/redmine/issues/11346

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -475,6 +475,8 @@ bundle edit_line ncf_insert_block(block) {
 
 # Sets the RHS of configuration items with a given separator
 # supports keys that contains special characters (like *, ?)
+# This bundle will allow any quantity of spaces before the separator, but none 
+# between the separator and the value
 bundle edit_line ncf_maintain_keys_values(v, sep) {
   vars:
       "index" slist => getindices("${v}");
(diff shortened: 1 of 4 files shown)


tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -494,7 +494,7 @@ bundle edit_line ncf_maintain_keys_values(v, sep) {
         expression => regline("^${keypat[${index}]}.*", "${edit.filename}");
 
   insert_lines:
-      "${index}${sep}$i{${v}[${index}]}"
```
ncharles: this worked because incorrect line was added, then corrected just below by the replace_patterns

tree/30_generic_methods/file_ensure_key_value_option.cf
```diff
@@ -0,0 +1,67 @@
+#####################################################################################
+# Copyright 2016 Normation SAS
```
amousset: 2017

tree/30_generic_methods/file_ensure_key_value_option.cf
```diff
@@ -0,0 +1,67 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File ensure key -> value present
```
amousset: We need to change the name

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
OK, merging this PR

PR #644, opened 2017-10-13
Fixes #11606: "Show deprecated methods" filter is broken

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11606

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -591,6 +591,7 @@ $scope.groupMethodsByCategory = function () {
   }
 
   $scope.checkFilterCategory = function(methods){
+    //If this function returns true, the category is displayed. Else, it is hidden by filters.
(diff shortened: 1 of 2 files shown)


builder/js/ncf.js
```diff
@@ -622,15 +622,17 @@ $scope.groupMethodsByCategory = function () {
     return method.agent_support.some(function(agent){return agent === agentSupport});
   }
   $scope.checkFilterMethod = function(method){
+    var deprecatedFilter = $scope.filter.showDeprecated || !method.deprecated;
```
VinceMacBuche: Can you add a comment at the begining of this function explaining what is expected ( return true => display the method, return false => filter the user)

RaphaelGauthier: Sure !!

RaphaelGauthier:
Commit modified

PR #645, opened 2017-10-17
Fixes #11601: Add a variable_string_from_command method

amousset:
https://www.rudder-project.org/redmine/issues/11601

### DIFF
--- a/tree/30_generic_methods/variable_string_from_command.cf
+++ b/tree/30_generic_methods/variable_string_from_command.cf
@@ -1,0 +1,96 @@
+#####################################################################################
+# Copyright 2017 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


amousset:
Do not merge for now, some small things to fix.

ncharles:
this looks ok, but it is scary
what is the small thing to fix ?

amousset:
Commit modified

amousset:
Executing commands without the full path failed because execresult though the `$()` was an undefined variable, because it is evaluated after replacing `const.dollar`.

I move the `/bin/true` call which was not actually necessary at the beginning inside the command, and added a specific test case.

tree/30_generic_methods/variable_string_from_command.cf
```diff
@@ -0,0 +1,93 @@
...
+      "promisers"          slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
+      "class_prefix"      string => canonify(join("_", "promisers"));
+      "args"               slist => { "${variable_prefix}", "${variable_name}", "${file_name}" };
+
+      # Why this ugly code? We need both the result and the return code of the command
+      # and we would prefer only executing it once.
+      #
+      # We concatenate the return code as a three char string, and extract it after.
+      #
+      # /bin/true is necessary because execresult thinks the $() is an unexpanded variable.
```
ncharles: can you update the doc to explain why the /bin/true is in the ${ } ?

peckpeck:
Looks good to me

amousset:
Commit modified

PR #651, opened 2017-11-14
Fixes #11729: eval usage is not compatible with CFEngine < 3.9

amousset:
https://www.rudder-project.org/redmine/issues/11729


amousset:
Commit modified

PR #652, opened 2017-11-15
Fixes #11732: Implement a push / pop mechanism for audit mode

peckpeck:
https://www.rudder-project.org/redmine/issues/11732

### DIFF
--- a/tree/20_cfe_basics/dry_run_lib.cf
+++ b/tree/20_cfe_basics/dry_run_lib.cf
@@ -35,6 +35,22 @@
 # When trying to enable dry-run on systems that do not support the dry-run lib,
 # the agent will abort execution to avoid making unwanted changes.
 #
+# There is a push/pop mechanism for dry-run mode.
+# This mechanism is meant to be used in generic methods when a specific mode is required temporarily
+# but the original mode must be restored afterwards.
+#
+# A real example should look like:
+#
+# ```
+# methods:
+#   "enable dry-run"  usebundle => push_dry_run_mode("true");
+#   "dry-run part"    usebundle => package_present("htop");
+#   "restore dry-run" usebundle => pop_dry_run_mode();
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/dry_run_lib.cf
```diff
@@ -82,6 +82,71 @@ bundle agent set_dry_run_mode(mode)
       "${configuration.debug} Enabling dry-run mode";
 }
 
+bundle agent dry_run_unique
```
amousset: Could you add a comment to explain is is used as a datastore by push/pop_dry_run_mode?

tree/20_cfe_basics/dry_run_lib.cf
```diff
@@ -82,6 +82,71 @@ bundle agent set_dry_run_mode(mode)
...
+{
+  vars:
+      "value" string => "1";
+      "stack" slist => {};
+}
+
```
amousset: Could you add a usage doc in a comment: how to initialize the stack and use it afterwards, what happens if you use set_dry_run_mode with oush/pop, etc.

peckpeck:
Commit modified

tree/20_cfe_basics/dry_run_lib.cf
```diff
@@ -82,6 +82,82 @@ bundle agent set_dry_run_mode(mode)
       "${configuration.debug} Enabling dry-run mode";
 }
 
+
+# Bundle that hold global values used by push_dry_run_mode and pop_dry_run_mode.
+# Do not call directly at this would reinitialize its content.
```
amousset: as

amousset:
I think we should also modify the documentation of the dry-run lib to advise using push/pop and give an example

peckpeck:
Commit modified

PR #656, opened 2017-11-27
Fixes #11792: Generic method file_ensure_key_value_present_in_ini_section.cf has leaky classes, resulting in invalid results

ncharles:
https://www.rudder-project.org/redmine/issues/11792

### DIFF
--- a/tests/acceptance/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
+++ b/tests/acceptance/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
@@ -57,7 +57,7 @@ ${base_text_up}";
     "name_3"            string => "name";
     "value_3"           string => "value";
     "new_value_3"       string => "newvalue";
-    "reference_3"       string => "${base_text_up}
+    "reference_3"       string => "[${section_3}]
 ${name_3}=${new_value_3}";
 
   commands:
@@ -88,7 +88,6 @@ ${name_3}=${new_value_3}";
   "/bin/echo"
       args    => "\"${name_3}=${value_3}\" > \"${file_3}\"",
       contain => in_shell;
-
 }
 
 #######################################################
--- a/tree/20_cfe_basics/cfengine/common.cf
+++ b/tree/20_cfe_basics/cfengine/common.cf
@@ -308,6 +308,102 @@ body classes classes_generic(x)
       promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
 }
 
+body classes results(scope, class_prefix)
+# @brief Define classes prefixed with `class_prefix` and suffixed with
+# appropriate outcomes: _kept, _repaired, _not_kept, _error, _failed,
+# _denied, _timeout, _reached
+#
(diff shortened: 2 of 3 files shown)


tests/acceptance/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
```diff
@@ -52,13 +52,14 @@ ${base_text_down}";
...
+# Third test: ensure that if the line already exists, it will be modified (must have an end section to work with cfengine <3.10)
     "section_3"         string => "section_test1";
     "name_3"            string => "name";
     "value_3"           string => "value";
     "new_value_3"       string => "newvalue";
-    "reference_3"       string => "${base_text_up}
-${name_3}=${new_value_3}";
+    "reference_3"       string => "[${section_3}]
+${name_3}=${new_value_3}
+[endfile]";
```
amousset: We should not require this, we have an exeption in stdlib to handle the behavior change.

ncharles: it does not work on cfengine 3.7 without this

ncharles: https://tracker.mender.io/browse/CFE-2263?jql=text%20~%20%22select_region%22

ncharles: it actually fails from 3.6.3 to 3.8 according to ticket

amousset: How could this test past before in this case?

ncharles: we only have agent in 3.6.2 and 3.10 if i'm correct

amousset: We have 3.6.5 in 3.1, and we had 3.7 in 3.2/4.0

ncharles: you are perfectly right, I was testing with a vanilla 3.7

ncharles:
Commit modified

tests/acceptance/30_generic_methods/file_ensure_key_value_present_in_ini_section.cf
```diff
@@ -57,7 +57,7 @@ ${base_text_up}";
     "name_3"            string => "name";
     "value_3"           string => "value";
     "new_value_3"       string => "newvalue";
-    "reference_3"       string => "${base_text_up}
+    "reference_3"       string => "[${section_3}]
```
amousset: why this change?

ncharles: for readability

ncharles: i scratched my head thrice over the test of why it was section1 in this test, so had to look up to base_text_up, which is not informative

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #659, opened 2017-11-30
Fixes #11816: Use built-in function and remove dumpdatastate usage in ncf

amousset:
https://www.rudder-project.org/redmine/issues/11816


amousset:
OK, merging this PR

PR #661, opened 2017-12-12
Fixes #11869: "Shared file to node" method does not send file to node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11869

### DIFF
--- a/tree/30_generic_methods/sharedfile_to_node.cf
+++ b/tree/30_generic_methods/sharedfile_to_node.cf
@@ -152,7 +152,7 @@ bundle agent sharedfile_to_node(target_uuid, file_id, file_path, ttl)
     _stdlib_path_exists_curl::
       "/opt/rudder/bin/rudder-sign"
         args => "${file_path} 1.1",
-        classes => classes_generic("${canonified_path}_signed"),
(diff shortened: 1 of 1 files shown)


ncharles:
this is dubious - it works out of the box without this patch on a 4.1

amousset:
this patch allows sending the file at the first run, without it I guess it works during the second run

VinceMacBuche:
It also allow to send the file if the file is created at every run, (i have a case where i want to send a file that is created by a command)

tree/30_generic_methods/sharedfile_to_node.cf
```diff
@@ -152,7 +152,7 @@ bundle agent sharedfile_to_node(target_uuid, file_id, file_path, ttl)
     _stdlib_path_exists_curl::
       "/opt/rudder/bin/rudder-sign"
         args => "${file_path} 1.1",
-        classes => classes_generic("${canonified_path}_signed"),
+        classes => if_ok("${canonified_path}_signed"),
```
ncharles: class is defined l 126: ${canonified_path}_signed" and => { fileexists("${file_path}.sign"), isnewerthan("${file_path}.sign","${file_path}") };

PR #669, opened 2017-12-17
Fixes #5673: Add support to parameters in ncf techniques written with ncf builder

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/5673

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -765,6 +765,7 @@ $scope.groupMethodsByCategory = function () {
     , "version": "1.0"
     , "bundle_name": undefined
     , "bundle_args": []
+    , "parameter": []
   };
 
   $scope.newTechnique = function() {
@@ -994,6 +995,12 @@ $scope.groupMethodsByCategory = function () {
     $scope.saving = false;
   });
 
+  $scope.newParam = {}
+  
+  $scope.addParameter = function() {
+    $scope.selectedTechnique.parameter.push(angular.copy($scope.newParam))
+    $scope.newParam.name = ""
+  }
   // Save a technique
   $scope.saveTechnique = function() {
     $scope.$broadcast('saving');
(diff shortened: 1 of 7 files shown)


builder/js/ncf.js
```diff
@@ -994,6 +995,15 @@ $scope.groupMethodsByCategory = function () {
     $scope.saving = false;
   });
 
+  $scope.newParam = {
+    "description" : "test"
+  }
+  $scope.addParameter = function() {
+    console.log($scope.newParam)
```
peckpeck: Debug info ?

peckpeck:
LGTM

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

builder/js/ncf.js
```diff
@@ -994,6 +995,12 @@ $scope.groupMethodsByCategory = function () {
     $scope.saving = false;
   });
 
+  $scope.newParam = {}
+  
+  $scope.addParameter = function() {
+    $scope.selectedTechnique.parameter.push(angular.copy($scope.newParam))
```
peckpeck: I don't understand how and when newParam is modified

tools/ncf.py
```diff
@@ -513,9 +522,20 @@ def generate_technique_content(technique, methods):
     # Add commentary for each new line in the metadata
     metadata_value = technique[metadata_key].replace("\n", "\n# ")
     content.append('# @'+ metadata_key +" "+ metadata_value)
+
+  bundle_param = ""
+  if technique["parameter"] > 0:
```
peckpeck: Why do you compare a list with an integer

tools/ncf.py
```diff
@@ -513,9 +522,20 @@ def generate_technique_content(technique, methods):
...
+    bundle_param = "("+", ".join([re.sub("\W","_",param["name"]) for param in technique["parameter"] ])+")"
```
peckpeck: Please use r"" or r'' insdeat of "" when writing a litteral regex.

peckpeck: Be careful with \W
It doesn't have the same behaviour depending on the locale and on the ancoding on the string.
And the default encoding of the string is not the same in python 2 and python 3

I think we already have a canoncalization function in python somewhere

builder/js/ncf.js
VinceMacBuche: In fact newParam does not need to exists anymore and could be replaced by an empty object to simplify things here

tools/ncf.py
VinceMacBuche: forgot a len ...

VinceMacBuche:
Commit modified

PR #671, opened 2017-12-20
Fixes #6291: Zypper hangs on dialogs / issues due to open tty

amousset:
https://www.rudder-project.org/redmine/issues/6291


amousset:
No idea why it was missing, tested on sles 10/11/12 without issues

PR #674, opened 2018-01-02
Fixes #11913: no reporting if value to report contains a variable

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11913

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -550,6 +550,7 @@ def generate_technique_content(technique, methods):
   for method_call in technique["method_calls"]:
     method_name = method_call["method_name"]
     method_info = methods[method_name]
+    # regex to match quote characters not preceded by a backslash
     regex = re.compile(r'(?<!\\)"', flags=re.UNICODE )
     # Treat each argument of the method_call
     if 'args' in method_call:
@@ -575,7 +576,15 @@ def generate_technique_content(technique, methods):
     else:
       promiser = "method_call"
 
-    content.append('    "'+promiser+'" usebundle => '+method_call['method_name']+'('+arg_value+'),')
+    # Set bundle context, first escape paramters
+    class_parameter_id    = method_info["class_parameter_id"] - 1
+    class_parameter_name  = regex.sub(r'\\"', method_info["parameter"][class_parameter_id]["name"])
+    class_parameter_value = regex.sub(r'\\"', method_call["args"][class_parameter_id])
+    technique_name        = regex.sub(r'\\"', technique["name"])
+    content.append('    "'+promiser+'_context" usebundle => _current_technique_report_info("'+technique_name+'", "'+class_parameter_name+'", "'+class_parameter_value+'");')
+
+    # Append method call
+    content.append('    "'+promiser+'" usebundle => '+method_name+'('+arg_value+'),')
     content.append('      ifvarclass => concat("'+class_context+'");')
 
   content.append('}')
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -29,211 +29,17 @@
 
 bundle agent log_rudder(message, old_class_prefix, origin_class_prefix, args)
 {
-  # The different files used for reporting are:
-  # * expected_reports_source: part of the promises, generated on the server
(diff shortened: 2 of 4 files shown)


tools/ncf.py
```diff
@@ -547,7 +547,11 @@ def generate_technique_content(technique, methods):
     else:
       promiser = "method_call"
 
-    content.append('    "'+promiser+'" usebundle => '+method_call['method_name']+'('+arg_value+'),')
+    class_parameter_id    = method_info["class_parameter_id"]
+    class_parameter_name  = method_info["parameter"][class_parameter_id]["name"]
+    class_parameter_value = method_call["args"][class_parameter_id]
+    content.append('    "'+promiser+'_context" usebundle => current_technique_report_info("'+method_info["name"]+'", "'+class_parameter_name+'", "'+class_parameter_value+'");')
```
ncharles: is it properly escaped ? what if it component key contains " or ' ?

peckpeck: This doesn't match current_technique_report_info call, the first parameter is technique_name

VinceMacBuche: Thanks @ncharles ! this needs escaping you are right :( :(

VinceMacBuche: Thanks @peckpeck  it should be the technique name not the method name ...

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

tools/ncf.py
```diff
@@ -550,6 +550,7 @@ def generate_technique_content(technique, methods):
   for method_call in technique["method_calls"]:
     method_name = method_call["method_name"]
     method_info = methods[method_name]
+    # regex to match quote characters not preceded by a backslash
```
ncharles: it seems to me it does a little more than backslash, or else the regexes lines 581-583 are not complete

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -29,212 +29,17 @@
...
-      # 5/ we aggregate this with the availability of the promiser stack to know if we should report using v2 format
-      "report_v2"           expression => "expected_report_v2_csv.expected_report_v2_res.method_v2.has_promiser_stack.class_prefix_v2";
-
-    any::
-      "no_old_class_prefix" expression => strcmp("${old_class_prefix}", "");
-
-      "pass3" expression => "pass2";
-      "pass2" expression => "pass1";
-      "pass1" expression => "any";
+    "c_class_prefix" string => canonify("${old_class_prefix}");
```
ncharles: could you add 2 spaces of indentation ?

VinceMacBuche: It's correctly idented, there is no class defined

<pre>
  vars:
    "c_class_prefix" ...
</pre>

tools/ncf.py
VinceMacBuche: No it's exactly that, the regex here in 554 is using a negative lookbehind: 

> (?<!text)

to match quotes not preceded by a backslash

then 581-583 use the regexp defined in 554 to replace the unescaped quotes by an escaped quote

tree/20_cfe_basics/log_rudder.cf
VinceMacBuche: or do indented - spacedces even iff there is no class ?

ncharles: yes, even if no class

VinceMacBuche: OK from what i read elsewhere we do likewise !! thanks @ncharles  !

tools/ncf.py
ncharles: My bad ! Thank you for the clarification

VinceMacBuche:
Commit modified

tree/20_cfe_basics/log_rudder.cf
VinceMacBuche: ok i reindented with 6 spaced (and the report below is also now indented with 6 spaces)

PR #675, opened 2018-01-10
Fixes #11937: yum is run at each cf-promises, having a massive perf cost, on redhat-like system

ncharles:
https://www.rudder-project.org/redmine/issues/11937

### DIFF
--- a/tree/20_cfe_basics/cfengine/packages.cf
+++ b/tree/20_cfe_basics/cfengine/packages.cf
@@ -115,7 +115,9 @@ bundle common redhat_no_locking_knowledge {
 
 
   classes:
-    redhat::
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/cfengine/packages.cf
```diff
@@ -115,7 +115,7 @@ bundle common redhat_no_locking_knowledge {
 
 
   classes:
-    redhat::
+    redhat.agent::
```
amousset: could you add comment here, the purpose of this class is not obvious

ncharles:
Commit modified

peckpeck:
@ncharles needs manual merging

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #678, opened 2018-01-16
Fixes #11894: Saving a technique in the technique editor in IE11 fails

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11894

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -7,7 +7,7 @@
         <link href="css/ngToast.min.css" rel="stylesheet">
         <link href="css/custom.css" rel="stylesheet">
         <!-- Needed polyfill for ie11, works in all other browsers -->
-        <script src="js/startswith.js"></script>
+        <script src="js/polyfill.js"></script>
         <script src="js/angular-1.3.14.min.js"></script>
         <script src="js/angular-messages-1.3.14.min.js"></script>
         <script src="js/angular-animate.min.js"></script>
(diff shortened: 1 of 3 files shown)


builder/index.html
```diff
@@ -7,7 +7,7 @@
         <link href="css/ngToast.min.css" rel="stylesheet">
         <link href="css/custom.css" rel="stylesheet">
         <!-- Needed polyfill for ie11, works in all other browsers -->
-        <script src="js/startswith.js"></script>
```
VinceMacBuche: Can you remove file startswith.js if it is not used anymore ??

builder/js/ncf.js
```diff
@@ -1014,8 +1015,11 @@ $scope.groupMethodsByCategory = function () {
         function(m) { 
           usedMethods.add($scope.generic_methods[m.method_name]);
         }
-      )
-      $http.post("/rudder/secure/api/latest/ncf",{"technique": ncfTechnique,"methods":Object.values($scope.generic_methods)});
+      );
+      var methodKeys = Object.keys($scope.generic_methods).map(function(e) {
```
VinceMacBuche: Why n o object.values ?

builder/index.html
VinceMacBuche: Ok saw the rename !! great

builder/js/ncf.js
RaphaelGauthier: Because IE11 doesn't support Object.values... :disappointed:

VinceMacBuche: Sadness :(

RaphaelGauthier:
OK, merging this PR

PR #680, opened 2018-01-16
Fixes #11974: Avoid waiting on lock when calling yum lost in 4.3

ncharles:
https://www.rudder-project.org/redmine/issues/11974

### DIFF
--- a/tree/20_cfe_basics/cfengine/packages.cf
+++ b/tree/20_cfe_basics/cfengine/packages.cf
@@ -257,6 +257,22 @@ bundle common rpm_knowledge
       "rpm3_arch_regex" string => "\S+\s+(\S+).*";
 }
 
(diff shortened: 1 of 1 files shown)


ncharles:
Commit modified

tree/20_cfe_basics/cfengine/packages.cf
```diff
@@ -257,6 +257,22 @@ bundle common rpm_knowledge
       "rpm3_arch_regex" string => "\S+\s+(\S+).*";
 }
```
amousset: could you move this bundle to ncf_lib.cf to make future sync easier?

nickanderson: Merged to master, backported to 3.7.x and 3.10.x.

Thank you!

PR #682, opened 2018-01-17
Fixes #11999: Wrong detection of package provider on debian for package_* method

ncharles:
https://www.rudder-project.org/redmine/issues/11999

### DIFF
--- a/tree/20_cfe_basics/package_lib.cf
+++ b/tree/20_cfe_basics/package_lib.cf
@@ -75,6 +75,8 @@ bundle agent ncf_package(name, version, architecture, provider, state, options)
 
       # Select the right package manager
       # The default one has already been selected if needed using mackage_module_knowledge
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/package_lib.cf
```diff
@@ -75,7 +75,7 @@ bundle agent ncf_package(name, version, architecture, provider, state, options)
 
       # Select the right package manager
       # The default one has already been selected if needed using mackage_module_knowledge
-      "use_apt_provider" expression => strcmp("apt", "${provider}");
+      "use_apt_provider" expression => strcmp("apt_get", "${provider}");
```
amousset: When specifying the provider in the generic method the parameter is `apt`, so we should detect both as valid `apt` provider.

ncharles: you are totally right ! thank you

ncharles:
Commit modified

PR #685, opened 2018-01-19
Fixes #12008: Is change message are mandatory, we cannot update technique with technique editor in Rudder 4.3

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12008


VinceMacBuche:
I'll redo it on 4.2 becasue it think the issue is there too

PR #686, opened 2018-01-19
Fixes #12008: If change message is mandatory, we cannot update technique with technique editor

VinceMacBuche:


VinceMacBuche:
Replacing #685

VinceMacBuche:
OK, merging this PR

PR #687, opened 2018-01-23
Fixes #12032: Rudder services are managed in a non deterministic way on Rudder 4.3 - Centos6

ncharles:
https://www.rudder-project.org/redmine/issues/12032

### DIFF
--- a/tree/30_generic_methods/service_ensure_started_at_boot.cf
+++ b/tree/30_generic_methods/service_ensure_started_at_boot.cf
@@ -41,7 +41,7 @@ bundle agent service_ensure_started_at_boot(service_name)
       "args"                     slist => { "${service_name}" };
 
   methods:
-    "action"             usebundle => service_stopped("${service_name}");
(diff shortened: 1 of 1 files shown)


ncharles:
Commit modified

tree/30_generic_methods/service_ensure_started_at_boot.cf
```diff
@@ -41,7 +41,7 @@ bundle agent service_ensure_started_at_boot(service_name)
       "args"                     slist => { "${service_name}" };
 
   methods:
-    "action"             usebundle => service_stopped("${service_name}");
+    "action"             usebundle => service_started("${service_name}");
```
amousset: service_enabled

ncharles: commit modified.
but now we have a duplicate between service_enabled and service_ensure_started_at_boot; how is  that useful ?

PR #693, opened 2018-02-01
Fixes #12078: ncf techniques are not reporting on the correct technique name

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12078


VinceMacBuche:
Commit modified

PR #696, opened 2018-02-02
Fixes #10304: Broken variable expression when using node properties in the technique editor

peckpeck:
https://www.rudder-project.org/redmine/issues/10304


peckpeck:
OK, merging this PR

PR #704, opened 2018-02-14
Fixes #11930: Add a way to prevent reporting of methods

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11930

### DIFF
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -99,6 +99,10 @@ bundle agent _rudder_common_report(technique_name, status, identifier, component
 #
 bundle agent _rudder_common_reports_generic(technique_name, class_prefix, identifier, component_name, component_key, message_prefix)
 {
+
+  class:
+      "report" expression => "${report_data.should_report}";
+
   methods:
 
       # This case should NEVER happen. If it ever happens, it is a bug in CFEngine or ncf that lead to changing something in dry-run mode.
@@ -107,7 +111,25 @@ bundle agent _rudder_common_reports_generic(technique_name, class_prefix, identi
                 action  => immediate_ignore_dry_run,
              ifvarclass => "(dry_run|global_dry_run).${class_prefix}_repaired";
 
-    !(dry_run|global_dry_run)::
+    !report::
+      "na"
+        usebundle  => _rudder_common_report("${technique_name}", "log_info", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} was not applicable"),
+        ifvarclass => "${class_prefix}_noop.!${class_prefix}_kept.!${class_prefix}_repaired.!${class_prefix}_error";
+
+      "success"
+        usebundle  => _rudder_common_report("${technique_name}", "log_info", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} was correct"),
+        ifvarclass => "${class_prefix}_kept.!${class_prefix}_repaired.!${class_prefix}_error";
+
+      "repaired"
+        usebundle  => _rudder_common_report("${technique_name}", "log_repaired", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} was repaired"),
+        ifvarclass => "${class_prefix}_repaired.!${class_prefix}_error";
+
+      "error"
+        usebundle  => _rudder_common_report("${technique_name}", "log_warn", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} could not be repaired"),
+        ifvarclass => "${class_prefix}_error";
+
+
+    report.!(dry_run|global_dry_run)::
       "na"
         usebundle  => _rudder_common_report("${technique_name}", "result_na", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} was not applicable"),
         ifvarclass => "${class_prefix}_noop.!${class_prefix}_kept.!${class_prefix}_repaired.!${class_prefix}_error";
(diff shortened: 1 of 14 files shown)


tree/20_cfe_basics/log_rudder.cf
```diff
@@ -202,3 +224,18 @@ bundle agent clean_reporting_context {
     "report_data.rule_id"        string => "";
     "report_data.identifier"     string => "";
 }
+
+bundle agent report_data {
+  vars:
+    "should_report" string => "true";
```
peckpeck: true and false are not classes in cfengine, use any and a class that is garanteed to be never defined

peckpeck: or use a cmp() in the class definition in generic methods

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -107,7 +111,25 @@ bundle agent _rudder_common_reports_generic(technique_name, class_prefix, identi
                 action  => immediate_ignore_dry_run,
              ifvarclass => "(dry_run|global_dry_run).${class_prefix}_repaired";
 
-    !(dry_run|global_dry_run)::
+    !report::
```
peckpeck: I don't understand the duplication of report calls

tree/20_cfe_basics/log_rudder.cf
VinceMacBuche: "true" and "false" can be used to defined classes like done in the other bundle

https://docs.cfengine.com/docs/3.10/reference-promise-types-classes.html#expression

peckpeck: your first url point to nothing.
My cfengine test code says it doesn't work:

  reports:
      "Done";
    true::
      "true";
    false::
      "false";

-> Done

VinceMacBuche: Here we are defining a variable not a class

VinceMacBuche: You should do likewise

VinceMacBuche: You should do this way:

vars:

  "should_report" string => "true";

classes:
  "report" expression => "${should_report}";

reports:

  report:: 
    "hey I have reporting based on ${should_report};

VinceMacBuche: bundle agent test {

vars:

  "should_report" string => "true";

classes:
  "report" expression => "true";

reports:

  "no  reporting based on ${should_report}";
  report::
  "hey I have reporting based on ${should_report}";
}

VinceMacBuche: here we only define the value "should_report"

VinceMacBuche: This works since cfengine 3.9

VinceMacBuche: All versions of Rudder (4.1 + are using cfengine 3.10)

VinceMacBuche: So we have to chose what we want to de here

VinceMacBuche:
PR rebased

VinceMacBuche:
@peckpeck If ncf renaming is merged, maybe we can merge this one. Attention: there is still some methods to modify but i think we should merge this first and male another pr to fix all others

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -107,7 +111,25 @@ bundle agent _rudder_common_reports_generic(technique_name, class_prefix, identi
...
+      "repaired"
+        usebundle  => _rudder_common_report("${technique_name}", "log_repaired", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} was repaired"),
+        ifvarclass => "${class_prefix}_repaired.!${class_prefix}_error";
+
+      "error"
+        usebundle  => _rudder_common_report("${technique_name}", "log_warn", "${identifier}", "${component_name}", "${component_key}", "${message_prefix} could not be repaired"),
+        ifvarclass => "${class_prefix}_error";
+
+
+    report&!(dry_run|global_dry_run)::
```
amousset: we usually use . for "and"

tree/30_generic_methods/service_ensure_running.cf
```diff
@@ -37,12 +37,14 @@ bundle agent service_ensure_running(service_name)
...
 
   methods:
+    "inner"     usebundle => calling_inner_method;
     "action"             usebundle => service_started("${service_name}");
     "class copy"         usebundle => _classes_copy("service_started_${canonified_service_name}", "${old_class_prefix}"),
                          ifvarclass => "service_started_${canonified_service_name}_reached";
     "new result classes" usebundle => _classes_copy("${class_prefix}_action", "${class_prefix}"),
                          ifvarclass => "${class_prefix}_action_reached";
     "report"
+    "end_inner"     usebundle => end_calling_inner_method;
```
amousset: the is inserted in the middle of a promise

VinceMacBuche:
Commit modified

tree/30_generic_methods/service_stop.cf
```diff
@@ -39,14 +39,25 @@ bundle agent service_stop(service_name)
     "class_prefix"      string => canonify(join("_", "promisers"));
     "args"                     slist => { "${service_name}" };
 
+  class:
+
+    "is_inner_method" expression => "!${report_data.should_report}";
+
   methods:
+    !is_inner_method::
```
amousset: why should service_action ever make reports here?

VinceMacBuche: In fact the condition is only needed on the end side we should always disable inner reporting, if it is already an inner method

VinceMacBuche:
Commit modified

tree/30_generic_methods/service_stop.cf
```diff
@@ -39,14 +39,24 @@ bundle agent service_stop(service_name)
     "class_prefix"      string => canonify(join("_", "promisers"));
     "args"                     slist => { "${service_name}" };
 
+  class:
+
+    "is_inner_method" expression => "!${report_data.should_report}";
```
peckpeck: Shouldn't we do that on ever method ? since any method can become a inner method

VinceMacBuche: Maybe, I suggest to do likewise in another PR :)

peckpeck: ok, in this case, this PR is OK for me

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #705, opened 2018-02-19
Fixes #11610: file_copy_with_check

amousset:
https://www.rudder-project.org/redmine/issues/11610

### DIFF
--- a/tests/acceptance/30_generic_methods/file_copy_from_local_source_with_check.cf
+++ b/tests/acceptance/30_generic_methods/file_copy_from_local_source_with_check.cf
@@ -1,0 +1,87 @@
+#######################################################
+#
+# Test checking if a file can be copied from a local source
+#
+#######################################################
+
+bundle common acc_path
+{
+  vars:
+    "root" string => getenv("NCF_TESTS_ACCEPTANCE", 1024);
+}
+
+body common control
+{
+      inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
+      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      version => "1.0";
+}
+
+#######################################################
+
+bundle agent init
+{
+  vars:
+    "tmp"                      string => getenv("TEMP", 1024);
+    "source_file"              string => "${tmp}/source_test";
+    "destination_file_1"       string => "${tmp}/destination_test1";
+    "destination_file_2"       string => "${tmp}/destination_test2";
+    "destination_file_3"       string => "${tmp}/destination_test3";
+    "destination_file_4"       string => "${tmp}/destination_test4";
+    "destination_file_canon_1" string => canonify("${destination_file_1}");
+    "destination_file_canon_2" string => canonify("${destination_file_2}");
+    "destination_file_canon_3" string => canonify("${destination_file_3}");
+    "destination_file_canon_4" string => canonify("${destination_file_4}");
+
+  files:
+    "${source_file}"
+      edit_line => insert_lines("test file"),
+      create => "true";
+    "${destination_file_3}"
+      edit_line => insert_lines("test file"),
+      create => "true";
+}
+
+#######################################################
+
+bundle agent test
+{
+  methods:
+    "ph1"  usebundle => file_copy_from_local_source_with_check("${init.source_file}", "${init.destination_file_1}", "/bin/true", "0");
+    "ph2"  usebundle => file_copy_from_local_source_with_check("${init.source_file}", "${init.destination_file_2}", "/bin/false", "0");
+    "ph3"  usebundle => file_copy_from_local_source_with_check("${init.source_file}", "${init.destination_file_3}", "/bin/false", "0");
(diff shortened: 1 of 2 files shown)


amousset:
I'll add a test with a different return code.

amousset:
Commit modified

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
```diff
@@ -0,0 +1,70 @@
...
+      "promisers"          slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
+      "class_prefix"      string => canonify(join("_", "promisers"));
+      "args"              slist => { "${source}", "${destination}", "${check_command}" };
+      "canonified_file"   string => canonify("${destination}");
+
+  defaults:
+      "rc_ok"             string => "0", if_match_regex => "";
+
+  classes:
+      "copy_needed" not => returnszero("diff \"${source}\" \"${destination}\"", "useshell");
```
ncharles: could you add an "agent:" class here, to make sure it's not run with cf-promises ?

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
```diff
@@ -0,0 +1,70 @@
...
+
+      "pass3"  expression => "pass2";
+      "pass2"  expression => "pass1";
+      "pass1"  expression => "any";
+
+  methods:
+      "check"  usebundle => condition_from_command("file_copy_from_local_source_with_check_${canonified_file}_check", "${check_command} \"${source}\"", "${rc_ok}", "999"),
```
ncharles: same here, we want the condition to be run only if the class agent is set (meaning, only if agent is running, not cf-promises, or any other ocmponent)
that would avoid triggering unwanted actions that might have side effet
(on a side note, it may be relevant to add this in the generic methods that runs actions also)

tests/acceptance/30_generic_methods/file_copy_from_local_source_with_check.cf
```diff
@@ -0,0 +1,78 @@
...
+    "destination_file_4"       string => "${tmp}/destination_test4";
+    "destination_file_canon_1" string => canonify("${destination_file_1}");
+    "destination_file_canon_2" string => canonify("${destination_file_2}");
+    "destination_file_canon_3" string => canonify("${destination_file_3}");
+    "destination_file_canon_4" string => canonify("${destination_file_4}");
+
+  files:
+    "${source_file}"
+      create => "true";
+    "${destination_file_3}"
```
ncharles: could you also do a test with non empty file ?

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
amousset: IIRC `commands` promises are not executed by cf-promises, only `returnszero` and `execresult`

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
amousset: yes!

amousset:
Commit modified

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
ncharles: you are totally right

amousset:
I'll add an example to the doc and a contraint to allow empty code

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
```diff
@@ -0,0 +1,67 @@
...
+      "class_prefix"      string => canonify(join("_", "promisers"));
+      "args"              slist => { "${source}", "${destination}", "${check_command}" };
+      "canonified_file"   string => canonify("${destination}");
+
+  defaults:
+      "rc_ok"             string => "0", if_match_regex => "";
+
+  classes:
+    agent::
+      "copy_needed" not => returnszero("diff \"${source}\" \"${destination}\"", "useshell");
```
ncharles: could you rather use ${paths.path[diff]} , in case PATH is not set

amousset:
Commit modified

tree/30_generic_methods/file_copy_from_local_source_with_check.cf
```diff
@@ -0,0 +1,86 @@
...
+# * If it is one of the `rc_ok` codes, it will copy `/tmp/program.conf` into `/etc/program.conf` and return a repaired
+# * If not, it will return an error
+#
+# @parameter source      Source file (absolute path on the target node)
+# @parameter destination Destination file (absolute path on the target node)
+# @parameter check_command   Command to run, it will get the source path as argument
+# @parameter rc_ok       Return codes to be considered as valid, separated by a comma (default is 0)
+# @parameter_constraint  rc_ok "allow_empty_string" : true
+# 
+# @class_prefix file_copy_from_local_source
```
ncharles: it should be file_copy_from_local_source_with_check

amousset:
Commit modified

amousset:
OK, merging this PR

amousset:
OK, merging this PR

PR #707, opened 2018-02-23
Fixes #12147: We cannot use file_enforce_content with a list and enforce=true

ncharles:
https://www.rudder-project.org/redmine/issues/12147


ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #708, opened 2018-02-26
Fixes #12154: Create a generic method to define environment variable

ncharles:
https://www.rudder-project.org/redmine/issues/12154

### DIFF
--- a/tree/30_generic_methods/environment_variable.cf
+++ b/tree/30_generic_methods/environment_variable.cf
@@ -1,0 +1,50 @@
+#####################################################################################
+# Copyright 2018 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


ncharles:
There's an issue with this solution: /etc/environment is not reloaded automatically by cf-execd (which is the process runnin cf-agent). So it won't be known by subsequent agent run, unless cf-execd is restarted

tree/30_generic_methods/environment_variable.cf
```diff
@@ -0,0 +1,49 @@
...
+# @description Enforce an environment variable value
+#
+# @parameter key Name of the environment variable
+# @parameter value Value of the environment variable
+# 
+# @class_prefix environment_variable
+# @class_parameter key
+# This bundle will define a class environment_variable_${key}_{kept,not_ok}
+
+bundle agent environment_variable(key, value)
```
amousset: environment_variable_present?

amousset: s/key/name/

peckpeck:
ping @ncharles

ncharles:
OK, merging this PR

PR #711, opened 2018-02-28
Fixes #12166: Rename some generic methods to match our naming convention - part3

amousset:
https://www.rudder-project.org/redmine/issues/12166


amousset:
OK, merging this PR

PR #712, opened 2018-02-28
Fixes #11857: Create hook ncf methods that can be used in pre-/post-run hooks

ncharles:
https://www.rudder-project.org/redmine/issues/11857

### DIFF
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -61,6 +61,7 @@ bundle agent log_rudder(message, old_class_prefix, origin_class_prefix, args)
 # component_name : the name of the component within the Technique
 # component_key  : the value of the component reference variable (or None if undefined)
 # message        : An explanation message understandable by a human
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/log_rudder.cf
```diff
@@ -142,6 +142,136 @@ bundle agent _rudder_common_reports_generic(technique_name, class_prefix, identi
         ifvarclass => "${class_prefix}_repaired";
 }
 
+
+######################################################################
+##### Reporting for the pre/post hooks
+##### These methods should not be used for normal generic method usage
+##### because they do not respect the stop on repair for audit mode
+######################################################################
```
amousset: Could you add a WARN on standard methods to notify these should be ketp in sync?

ncharles:
Commit modified

ncharles:
PR rebased

tree/20_cfe_basics/log_rudder.cf
ncharles: Done

PR #719, opened 2018-03-23
Fixes #12227: Migrate generic methods towards new loging system

peckpeck:
https://www.rudder-project.org/redmine/issues/12227

### DIFF
--- a/tree/30_generic_methods/_service_check_running_smf.cf
+++ b/tree/30_generic_methods/_service_check_running_smf.cf
@@ -36,14 +36,19 @@ bundle agent _service_check_running_smf(service_name)
       "class_prefix"        string => canonify(join("_", "promisers"));
       "args"                 slist => { "${service_name}" };
 
-  methods:
 
+  classes:
+      "should_report"    expression => "${report_data.should_report}";
+
+  methods:
       "disable_reporting"     usebundle => disable_reporting;
+
 
       "check_if_service_is_running" usebundle => service_check_running("${service_name}");
 
-      "enable_reporting" usebundle => enable_reporting;
 
+      "reenable_reporting"    usebundle => enable_reporting,
+                             ifvarclass => "should_report";
       "report"
         usebundle  => _log("Check if the service ${service_name} is started using SMF", "${old_class_prefix}", "${class_prefix}", @{args});
 }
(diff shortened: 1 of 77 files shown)


tree/30_generic_methods/_service_check_running_smf.cf
```diff
@@ -36,14 +36,19 @@ bundle agent _service_check_running_smf(service_name)
...
+  methods:
       "disable_reporting"     usebundle => disable_reporting;
 
+
       "check_if_service_is_running" usebundle => service_check_running("${service_name}");
 
-      "enable_reporting" usebundle => enable_reporting;
 
+      "reenable_reporting"    usebundle => enable_reporting,
+                             ifvarclass => "report_mode";
```
VinceMacBuche: You should use if instead of ifvarclass

VinceMacBuche: More details here https://github.com/Normation/ncf/pull/720

VinceMacBuche: @peckpeck You should read that: https://en.wikipedia.org/wiki/Broken_windows_theory

tree/30_generic_methods/_service_check_running_smf.cf
```diff
@@ -36,14 +36,19 @@ bundle agent _service_check_running_smf(service_name)
       "class_prefix"        string => canonify(join("_", "promisers"));
       "args"                 slist => { "${service_name}" };
 
-  methods:
 
+  classes:
+      "report_mode"    expression => "${report_data.should_report}";
```
amousset: Could the name of the class be boolean like `should_report` or `reports_enabled`? report_mode means nothing.

ncharles: Caution, we have "full_compliance", "changes_only" and "reports_disabled" asglobal classes for reporting - we should not use a name too close to these naming, to avoid us major headache.

ncharles: should_report sounds correct - it matches the name of the variable, so it's easily grepable

peckpeck:
PR rebased

peckpeck:
Commit modified

PR #724, opened 2018-04-05
Fixes #12371: Some methods which let another one do the reporting are broken in 4.3

amousset:
https://www.rudder-project.org/redmine/issues/12371


amousset:
Do not merge for now, this does not seem to work correctly

amousset:
Commit modified

PR #725, opened 2018-04-05
Fixes #12368: Unexpected report on several services methods

amousset:
https://www.rudder-project.org/redmine/issues/12368


amousset:
Commit modified

amousset:
Commit modified

PR #727, opened 2018-04-05
Fixes #12378: Job ids in job methods should be canonified

amousset:
https://www.rudder-project.org/redmine/issues/12378

### DIFF
--- a/tree/30_generic_methods/schedule_simple_catchup.cf
+++ b/tree/30_generic_methods/schedule_simple_catchup.cf
@@ -37,7 +37,6 @@
 # @parameter periodicity_minutes   How often should the job run
 # @parameter periodicity_hours     How often should the job run
 # @parameter periodicity_days      How often should the job run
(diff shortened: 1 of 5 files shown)


amousset:
Commit modified

tree/30_generic_methods/schedule_simple_catchup.cf
```diff
@@ -37,7 +37,6 @@
 # @parameter periodicity_minutes   How often should the job run
 # @parameter periodicity_hours     How often should the job run
 # @parameter periodicity_days      How often should the job run
-# @parameter mode                  "nodups": avoid duplicate runs in the same period / "catchup": avoid duplicates and one or more run have been missed, run once before next period / "stateless": no check is done on past runs
```
amousset: This is a hard incompatibility that may break users policy, but as this method always broke policies when used from the technique editor, it should not be an issue...

peckpeck: this parameter should not be there 
And it never worked from the technique editor, so tyhis is ok

amousset:
Commit modified

amousset:
OK, merging this PR

PR #728, opened 2018-04-06
Fixes #12379: Jinja2 templating: Shows error in output but does not do Error report

amousset:
https://www.rudder-project.org/redmine/issues/12379


amousset:
OK, merging this PR

PR #732, opened 2018-04-10
Fixes #12392: Parent task broke policies

amousset:
https://www.rudder-project.org/redmine/issues/12392


amousset:
OK, merging this PR

PR #733, opened 2018-04-10
Fixes #12398: When upgrading from 4.1 to 4.3, reports from Rudder Techniques break reporting

ncharles:
https://www.rudder-project.org/redmine/issues/12398


amousset:
`defaults` are between `vars` and `classes`. Don't we also need to do it on the same bundles in rudder-techniques?

PR #734, opened 2018-04-11
Fixes #12414: Make reporting work when targeting several time the same item with several generic method

ncharles:
https://www.rudder-project.org/redmine/issues/12414


tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -879,14 +879,14 @@ body classes classes_generic_return_boolean_list_two(x, y, boolean_prefix, true_
...
-  !has_promiser_stack::
+#  !has_promiser_stack::
     promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
     repair_failed => { "repair_failed_$(x)", "$(x)_failed", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
     repair_denied => { "repair_denied_$(x)", "$(x)_denied", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
     repair_timeout => { "repair_timeout_$(x)", "$(x)_timeout", "$(x)_not_ok", "$(x)_error", "$(x)_not_kept", "$(x)_not_repaired", "$(x)_reached" };
     promise_kept => { "promise_kept_$(x)", "$(x)_kept", "$(x)_ok", "$(x)_not_repaired", "$(x)_reached" };
 
-  has_promiser_stack::
+#  has_promiser_stack::
```
ncharles: we should drop has_promiser_stack, and maybe define a class in the generic method that advertise that it use the class_prefix

ncharles:
Commit modified

amousset:
This is great!

tree/30_generic_methods/file_key_value_present_option.cf
```diff
@@ -36,10 +36,11 @@ bundle agent file_key_value_present_option(file, key, value, separator, option)
 {
   vars:
       "old_class_prefix" string => canonify("file_key_value_present_${file}");
-      "promisers"          slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
-      "class_prefix"      string => canonify(join("_", "promisers"));
       "args"              slist => { "${file}", "${key}", "${value}","${separator}", "${option}" };
 
+      "report_param"      string => join("_", args);
+      "class_prefix"      string => canonify("${old_class_prefix}_${report_param}");
```
amousset: This may easily break if this string gets longer than 1024 chars

amousset: I think it fails clearly though (and could could check this in the technique editor)

amousset: This could fail if string is longer than 1024 chars.

ncharles:
Commit modified

ncharles:
Commit modified

tree/30_generic_methods/file_key_value_present_option.cf
ncharles: this fails nicely: classes are not defined. However, we need in this case to catch the fact that classes are too long (see log_rudder.cf)

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -879,14 +879,14 @@ body classes classes_generic_return_boolean_list_two(x, y, boolean_prefix, true_
 # Define x and y prefixed/suffixed with promise outcome
 body classes classes_generic_two(x,y)
 {
-  !has_promiser_stack::
+#  !has_promiser_stack::
     promise_repaired => { "promise_repaired_$(x)", "$(x)_repaired", "$(x)_ok", "$(x)_reached" };
```
peckpeck: You should drop those lines too since they duplicate what is done with has_promiser_stack below

ncharles: you are totally right !

ncharles:
Commit modified

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -32,13 +32,34 @@ bundle agent log_rudder(message, old_class_prefix, origin_class_prefix, args)
```
amousset: could you rename the parameters to make their meaning clearer?

ncharles:
Generic methods code have been replaced with
for i  in `grep -L "disable_reporting" [a-z]*.cf`; do sed -i "/promisers/d" $i ;   done
for i  in `grep -L "disable_reporting" [a-z]*.cf`; do sed -i '/.*"args".*slist.*/a\      "report_param"      string => join("_", args);' $i ;   done
for i  in `grep -L "disable_reporting" [a-z]*.cf`; do sed -i '/.*join("_", args.*/a\      "class_prefix"      string => canonify("${old_class_prefix}_${report_param}");' $i ;   done

ncharles:
Commit modified

ncharles:
Commit modified

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -32,13 +32,29 @@ bundle agent log_rudder(message, old_class_prefix, origin_class_prefix, args)
...
 
-  methods:
+      "class_prefix_length"   int => string_length("${origin_class_prefix}");
+
+  classes:
+      # Use the full class prefix if:
+      # - it is defined
+      # - its length is smaller than 1000 character (classes are limited to 1023 characters, so if the prefix
+      #   is too long, then classes are not defined and it can't be used
+      # Note that if we don't use the full class_prefix, we fallback to the previous use case, where classes collision happen
```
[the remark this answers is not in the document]
ncharles: done

tests/acceptance/30_generic_methods/log_rudder.cf
```diff
@@ -13,7 +13,7 @@ bundle common acc_path
 body common control
 {
       inputs => { "${acc_path.root}/default.cf.sub", "${acc_path.root}/default_ncf.cf.sub", "@{ncf_inputs.default_files}" };
-      bundlesequence  => { configuration, default("${this.promise_filename}") };
+      bundlesequence  => { configuration, default("$(this.promise_filename)") };
```
amousset: why changing this?

tests/acceptance/30_generic_methods/log_rudder.cf
```diff
@@ -1,8 +1,8 @@
-#######################################################
+#########################################################################################
 #
-# Test checking if reporting is correcly done with rudder_logger
+# This will check reporting within log_rudder, by checking the defined classes
```
amousset: Could you keep this test and make the class test a new one? It is good to keep a test of the reports output too.

ncharles: Damn - i overwrote by mistake the test :/ i'll correct this ASAP

ncharles:
Commit modified

PR #743, opened 2018-04-16
Fixes #12451: In some methods reporting is broken because class_prefix is not correctly updated when using another method 

amousset:
https://www.rudder-project.org/redmine/issues/12451


amousset:
OK, merging this PR

PR #761, opened 2018-05-24
Fixes #12597: Do not modify variables in existing bundles

amousset:
https://www.rudder-project.org/redmine/issues/12597


amousset:
Commit modified

amousset:
PR rebased

PR #765, opened 2018-05-29
Fixes #12693: Reporting context is not set correctly when using an iterator somewhere

ncharles:
https://www.rudder-project.org/redmine/issues/12693


tree/30_generic_methods/_file_check_special.cf
```diff
@@ -61,5 +61,5 @@ bundle agent file_check_special(file_name, file_type)
 
     any::
       "report"
-        usebundle => _log("Check if ${file_name} exists and is a ${file_type}", "${old_class_prefix}", "${class_prefix}", @{args});
+        usebundle => _log_v3("Check if ${file_name} exists and is a ${file_type}", "${file_name }", "${old_class_prefix}", "${class_prefix}", @{args});
```
peckpeck: there is a space in the variable name

ncharles: good catch ! Thank you

ncharles:
Commit modified

PR #769, opened 2018-06-07
Fixes #12761: Load custom jinja2 filters and tests for jinja2 templating

itacos:
https://www.rudder-project.org/redmine/issues/12761

### DIFF
--- a/tree/10_ncf_internals/modules/templates/jinja2-templating.py
+++ b/tree/10_ncf_internals/modules/templates/jinja2-templating.py
@@ -24,8 +24,10 @@ from optparse import OptionParser
 
 import jinja2
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
+import jinja2.loaders
 
 from distutils.version import StrictVersion
+import pkgutil, importlib
 
 try:
     import simplejson as json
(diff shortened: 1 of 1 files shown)


itacos:
Load custom Jinja2 filters and tests from `jinja2_custom` module in the /var/rudder/configuration-repository/ncf/modules/templates/ directory.

The filters and tests to be loaded must be listed in two dictionaries,
named  `FILTERS` and `TESTS` respectively. E.g.:

```python
   def comment(t):
      return '# '+t

   def is_odd(n):
      return True if (n % 2) else False

   FILTERS = {'comment': comment,}
   TESTS = {'is_odd': is_odd,}
```

tree/10_ncf_internals/modules/templates/jinja2-templating.py
```diff
@@ -66,6 +68,24 @@ def render(opts, args):
     if opts.strict:
         env.undefined = StrictUndefined
 
+    # Register customs
+    sys.path.append('/var/rudder/ncf/local/modules/templates/')
```
ncharles: the full path for the local template should be  /var/rudder/ncf/common/10_ncf_internals/modules/templates

tree/10_ncf_internals/modules/templates/jinja2-templating.py
```diff
@@ -24,8 +24,10 @@
 
 import jinja2
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
+import jinja2.loaders
```
ncharles: i'm not expert on jinja2, does is need an extra dependency ?

itacos: You are right, this import is not necessary. I will remove it.

tree/10_ncf_internals/modules/templates/jinja2-templating.py
itacos: You are right but the path provided at line 72 is the path of the custom filter module (aka `jinja2_custom`) and it must be outside of any overwritten mechanism.

itacos: To be copied to /var/rudder/ncf/local/modules/templates/ you want to say ?

ncharles: yes, i made a typo, sorry

PR #778, opened 2018-06-20
Fixes #12127: Backport options for yum package provider

amousset:
https://www.rudder-project.org/redmine/issues/12127


amousset:
OK, merging this PR

PR #787, opened 2018-07-04
Fixes #12887: fail to restart certain services on Ubuntu because of incomplete detection of systemd/upstart

ncharles:
https://www.rudder-project.org/redmine/issues/12887


ncharles:
OK, merging this PR

PR #793, opened 2018-07-16
Fixes #12970: Broken technique editor if a technique is in the browser cache, but has been removed from the FS

ncharles:
https://www.rudder-project.org/redmine/issues/12970


ncharles:
OK, merging this PR

PR #802, opened 2018-08-13
Fixes #13155: Huge agent logs when we define an new class_prefix that is longer than 1024 chars

ncharles:
https://www.rudder-project.org/redmine/issues/13155


ncharles:
all those that have a disable_reporting have been skipped - 

code for updating is : 
`
 for i in `grep -L "disable_reporting" [a-z]*.cf`; do sed -i 's/"class_prefix"      string/"full_class_prefix" string/'  $i ; done
for i in `grep -L "disable_reporting" [a-z]*.cf`; do sed -i '/"full_class_prefix/a\      "class_prefix"      string => string_head("${full_class_prefix}", "1000");' $i ; done
`

all those with a disable_reporting need to be fixed in https://www.rudder-project.org/redmine/issues/13202 - and this need probably a bit of thinking

ncharles:
Commit modified

PR #803, opened 2018-08-14
Fixes #13203: fix reporting in deprecated files methods in rudder 4.3 to support multi-reporting

ncharles:
https://www.rudder-project.org/redmine/issues/13203


ncharles:
i skipped file_content, as it has behaviour with list and without list, and i'm not sure how to manage that properly

ncharles:
there are some massive issue in file_content + file_enforce_content

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
Commit modified

tree/30_generic_methods/_file_check_special.cf
```diff
@@ -35,10 +35,11 @@ bundle agent file_check_special(file_name, file_type)
       "file_has_type"       string => filestat("${file_name}", "type");
 
       "old_class_prefix"    string => canonify("file_check_${file_type}_${file_name}");
-      "promisers"            slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
-      "class_prefix"        string => canonify(join("_", "promisers"));
-      "args"                slist => { "${file_name}", "${file_type}" };
 
+      "args"                 slist => { "${file_name}", "${file_type}" };
+      "report_param"        string => join("_", args);
```
amousset: Couldn't we have a problem if the sum of variables length is more than max var size? Shouldn't we stay compatible with old_class prefix everywhere?

tree/30_generic_methods/condition_from_command.cf
```diff
@@ -67,7 +67,7 @@ bundle agent condition_from_command(condition_prefix, command, true_codes, false
   methods:
       "report"
         usebundle  => _log_v3("Execute the test command ${command} to create ${condition_prefix}_{true,false}", "${condition_prefix}", "${old_class_prefix}", "${class_prefix}", @{args}),
-        ifvarclass => "(!has_promiser_stack.${old_class_prefix}_reached)|(has_promiser_stack.${class_prefix}_reached)";
+        ifvarclass => "${class_prefix}_reached";
```
amousset: For example in this case, class_prefix may not be defined if command length is almost 4k.

tree/30_generic_methods/_file_check_special.cf
ncharles: That's an interesting remark.

I tried several generic method with large string, and got failure elsewhere (when running command for instance), but the strings related to classes seem to be correctly truncated when concatenated like that (i only got a crash when I inputed a string larger than 16384 char in the vars: "bla" string => "here" itseld

ncharles: i can't see how we can detect that class_prefix has an issue, and if so, fallback to old_class_prefix

ncharles:
Commit modified

tree/30_generic_methods/condition_from_command.cf
ncharles: I just added an unit test for a 4k command

tree/30_generic_methods/_file_check_special.cf
amousset: it seems to work well indeed, there is no hard limit on var size, but only in the parser.

tree/30_generic_methods/file_ensure_lines_present.cf
```diff
@@ -30,25 +30,30 @@
...
 {
   vars:
-      "old_class_prefix" string => canonify("file_ensure_lines_present_${file}");
-      "promisers"          slist => { @{this.callers_promisers}, cf_null }, policy => "ifdefined";
-      "class_prefix"      string => canonify(join("_", "promisers"));
+      "old_class_prefix"  string => canonify("file_ensure_lines_present_${file}");
+
       "args"               slist => { "${file}", "${lines}" };
+      "report_param"      string => join("_", args);
+      "full_class_prefix" string => canonify("file_lines_present_${report_param}");
```
amousset: shouldn't it be `file_ensure_lines_present`?

ncharles: you are totally right

ncharles:
Commit modified

PR #811, opened 2018-08-21
Fixes #13249: Some ncf tests are too old and fails on many distributions

Fdall:
https://www.rudder-project.org/redmine/issues/13249


Fdall:
Commit modified

PR #816, opened 2018-08-23
Fixes #13277: Makefile fails if a symbolink link /usr/bin/ncf already exists

Fdall:
https://www.rudder-project.org/redmine/issues/13277


Fdall:
Commit modified

PR #826, opened 2018-09-19
Fixes #13523: variable_string and variable_string_default GM are failing in rudder 5.0

Fdall:
https://www.rudder-project.org/redmine/issues/13523


amousset:
Could you fix it in previous branches too, so we can test older ncf with new agents?

Fdall:
PR replaced by https://github.com/Normation/ncf/pull/828

PR #832, opened 2018-09-21
Fixes #13553: Test generic method output

amousset:
https://www.rudder-project.org/redmine/issues/13553


amousset:
We need to:
* Initialize a specific ncf.conf in the workdir
* Only look for reports in file in 30_generic_methods, that do not start by an underscore

amousset:
Commit modified

amousset:
Commit modified

PR #836, opened 2018-10-04
Update file_ensure_lines_present.cf

VinceMacBuche:


amousset:
I think it may break reporting with several editions on the same file as the reports only differ by the inserted content, and some of them might not be done is the message is exactly the same.

VinceMacBuche:
we could add the lines in the promiser

VinceMacBuche:
or generate id on the promiser

VinceMacBuche:
Done in #837

PR #837, opened 2018-10-04
Fixes #13608: 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/13608


peckpeck:
Won't we loose some logs with this solution ?

VinceMacBuche:
Yes, we talked about it last week .... We need to do something else to force reporting

VinceMacBuche:
put all variables in promiser or generate an id in all methods

VinceMacBuche:
and use it everywhere and do it log too

VinceMacBuche:
But I think we need to work with @amousset on this

VinceMacBuche:
PR rebased

PR #850, opened 2018-10-17
Fixes #13669: Missing status when I use `File copy from Rudder shared folder`

Fdall:
https://www.rudder-project.org/redmine/issues/13669

### DIFF
--- a/tree/30_generic_methods/file_from_shared_folder.cf
+++ b/tree/30_generic_methods/file_from_shared_folder.cf
@@ -34,7 +34,7 @@ bundle agent file_from_shared_folder(source, destination, hash_type)
 
       "old_class_prefix"        string => "file_from_shared_folder_${canonified_destination}";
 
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/file_from_shared_folder.cf
```diff
@@ -36,7 +36,7 @@ bundle agent file_from_shared_folder(source, destination, hash_type)
 
       "args"                     slist => { "${source}", "${destination}" };
       "report_param"            string => join("_", args);
-      "full_class_prefix"       string => canonify("file_from_shared_folder_${report_param}");
+      "full_class_prefix"       string => canonify("file_from_shared_folder_${canonified_destination}");
```
ncharles: this is breaking the n reports for a generic ethod, you should not do that

Fdall:
Commit modified

Fdall:
Commit modified

PR #851, opened 2018-10-18
Fixes #13683: Broken reporting in permisisons recurse

amousset:
https://www.rudder-project.org/redmine/issues/13683


amousset:
Not tested.

PR #862, opened 2018-10-30
Fixes #13745: change the detection of curl path in ncf to use rudder_curl if available

ncharles:
https://www.rudder-project.org/redmine/issues/13745

### DIFF
--- a/tree/20_cfe_basics/ncf_lib.cf
+++ b/tree/20_cfe_basics/ncf_lib.cf
@@ -22,6 +22,27 @@ bundle common ncf_const {
       "s" string => " ";
 }
 
+
+# Use rudder_curl variable if available, rather than the path[curl]
+bundle common ncf_paths {
+  vars:
+      "path[curl]"   string => "${g.rudder_curl}",
+                         if => isvariable("g.rudder_curl");
+
+      "path[curl]"   string => "${paths.path[curl]}",
+                     unless => isvariable("g.rudder_curl");
+
+     
+      "all_paths"     slist => getindices("path");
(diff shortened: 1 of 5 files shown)


tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -22,6 +22,20 @@ bundle common ncf_const {
       "s" string => " ";
 }
 
+
+# Use rudder_curl variable if available, rather than the path[curl]
+bundle common ncf_path {
+  vars:
+      "ncf_curl" string => "${g.rudder_curl}",
```
amousset: Could you put it in a `path` array for consistency with stdlib?

amousset: And name the bundle `ncf_paths`?

amousset: BTW I like this idea of overriding stdlib in ncf :+1:

ncharles:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -22,6 +22,27 @@ bundle common ncf_const {
...
+bundle common ncf_paths {
+  vars:
+      "path[curl]"   string => "${g.rudder_curl}",
+                         if => isvariable("g.rudder_curl");
+
+      "path[curl]"   string => "${paths.path[curl]}",
+                     unless => isvariable("g.rudder_curl");
+
+     
+      "all_paths"     slist => getindices("path");
```
amousset: :+1:

PR #866, opened 2018-10-31
Fixes #13753: sharedfile_to_node is always repaired

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/13753

### DIFF
--- a/tree/30_generic_methods/sharedfile_to_node.cf
+++ b/tree/30_generic_methods/sharedfile_to_node.cf
@@ -159,7 +159,7 @@ bundle agent sharedfile_to_node(target_uuid, file_id, file_path, ttl)
         ifvarclass => "!${canonified_path}_signed";
 
     _stdlib_path_exists_curl.pass2::
-      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || [ $code -eq 404 ] && exit 22 || exit 254"
+      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || if [ $code -eq 404 ]; then exit 22; elif [ $code -eq 200 ]; then exit 0; else exit 254; fi"
         contain => in_shell,
         classes => classes_generic_return_single_code_two("${old_class_prefix}_exist", "${class_prefix}_exist", "0", "22", "254"), # 254 = http 400 or above, 22 = http 404
         ifvarclass => "${canonified_path}_signed";


tree/30_generic_methods/sharedfile_to_node.cf
```diff
@@ -159,7 +159,7 @@ bundle agent sharedfile_to_node(target_uuid, file_id, file_path, ttl)
         ifvarclass => "!${canonified_path}_signed";
 
     _stdlib_path_exists_curl.pass2::
-      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || [ $code -eq 404 ] && exit 22 || exit 254"
+      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || if [ $code -eq 404 ]; then exit 22; else if [ $code -eq 200 ]; then exit 0; else exit 254; fi; fi"
```
peckpeck: else if -> elif

VinceMacBuche:
Commit modified

tree/30_generic_methods/sharedfile_to_node.cf
```diff
@@ -159,7 +159,7 @@ bundle agent sharedfile_to_node(target_uuid, file_id, file_path, ttl)
         ifvarclass => "!${canonified_path}_signed";
 
     _stdlib_path_exists_curl.pass2::
-      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || [ $code -eq 404 ] && exit 22 || exit 254"
+      "code=`${paths.path[curl]} ${url}?hash=${hash} --insecure --tlsv1 --location --head --output /dev/null --silent --fail --write-out %{http_code}` || if [ $code -eq 404 ]; then exit 22; elif [ $code -eq 200 ]; then exit 0; else exit 254; fi"
         contain => in_shell,
         classes => classes_generic_return_single_code_two("${old_class_prefix}_exist", "${class_prefix}_exist", "0", "22", "254"), # 254 = http 400 or above, 22 = http 404
```
peckpeck: This has changed : 404 -> 22, 200 -> 0, 254 otherwise
This also means that 20X and 30X are considered as errors, is it what we want ?

VinceMacBuche: That's documented likewise we could also change it to manage other codes. But I think in another Pull request

VinceMacBuche: 404 -> 22, 200 -> 0, 254 otherwise is what we expect

VinceMacBuche: But we could 404 -> 22, 400+ => 254, 0 for all other case but this may allow to much "success"

VinceMacBuche: Yes, the behavior is now correct which is better is supose ;)

VinceMacBuche:
OK, merging this PR

PR #872, opened 2018-11-12
Fixes #13562: file_keys_values_present does not suport some separators

peckpeck:
https://issues.rudder.io/issues/13562


peckpeck:
Commit modified

tree/20_cfe_basics/ncf_lib.cf
```diff
@@ -491,9 +491,10 @@ bundle edit_line ncf_maintain_keys_values(v, sep) {
       # Keys and values may contain regexps. Escape them for replace_pattern matching.
       "ve[${index}]" string => escape("${${v}[${index}]}");
       "ke[${index}]" string => escape("${index}");
+      "separator"    string => escape("${sep}");
```
amousset: line 505 should be fixed too

amousset: 550 too

ncharles: insert_lines should not have the escape

PR #884, opened 2018-12-17
Fixes #13984: Add a bundle to define persistent classes for job scheduler techniques

ncharles:
https://issues.rudder.io/issues/13984


ncharles:
to be merged before https://github.com/Normation/rudder-techniques/pull/1370

PR #887, opened 2018-12-18
Fixes #14000: Move generic stuff from techniques into ncf

amousset:
https://issues.rudder.io/issues/14000


amousset:
Commit modified

PR #894, opened 2019-01-09
Fixes #14125: Move stuff from rudder-lib into ncf

amousset:
https://issues.rudder.io/issues/14125


amousset:
OK, merging this PR

PR #900, opened 2019-01-16
Fixes #14179: Jinja templating from the doc fails everytime

Fdall:
https://issues.rudder.io/issues/14179

### DIFF
--- a/tree/30_generic_methods/file_from_template_jinja2.cf
+++ b/tree/30_generic_methods/file_from_template_jinja2.cf
@@ -177,7 +177,7 @@
 #
 #
 # # will display the current properties key/value pair
-# {% for key, value in vars.variable_prefix.dict_name.properties %}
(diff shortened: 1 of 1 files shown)


tree/30_generic_methods/file_from_template_jinja2.cf
```diff
@@ -161,23 +161,23 @@
 # Use the following template:
 #
 # ```jinja2
-# {% for item in vars.variable_prefix.dict_name.hosts %}
+# {% for item in vars.my_prefix.my_name.hosts %}
```
amousset: why renaming the variable?

Fdall: My bad, I had at first a missleading error, and though that the name "variable_prefix" was already in use. I double checked and there is no need to rename the variables.

Fdall:
Commit modified

PR #915, opened 2019-01-23
Fixes #14225: Enable 4h cache on available upgrades list

amousset:
https://issues.rudder.io/issues/14225


amousset:
OK, merging this PR

PR #934, opened 2019-03-05
Fixes #14390: Missing report on generic method "command exec" with $()

ncharles:
https://issues.rudder.io/issues/14390

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -35,6 +35,16 @@ function findIndex(array, elem) {
     }
     return -1;
 };
+// check if the parameter used contains $(somecontent), where somecontent is neither
+// a valid variable name (key.value), nor does contain /
+// If the content is invalid, returns true, else false
+var re = new RegExp(/\$\([^./]*\)/, 'm');
+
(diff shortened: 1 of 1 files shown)


builder/js/ncf.js
```diff
@@ -1149,7 +1159,30 @@ $scope.onImportFileChange = function (fileEl) {
           // We will lose the link between the selected method and the technique, to prevent unintended behavior, close the edit method panel
           $scope.selectedMethod = undefined;
         }
-        $scope.resetFlags();
+        $scope.resetFlags();i
```
VinceMacBuche: There is a 'i' after the semi-colon here!!

amousset: :eye: :eye_speech_bubble: :eyes: :bird:

ncharles: JS is so forgiving

ncharles:
Commit modified

builder/js/ncf.js
VinceMacBuche: i does nothing, you only call a variables that is undefined ... but yes ...

PR #961, opened 2019-04-12
Fixes #9130: Reporting missing if several generic methods with same class_prefix and different condition applied 

ncharles:
https://issues.rudder.io/issues/9130


ncharles:
Commit modified

PR #988, opened 2019-06-20
Fixes #15110: We should test sles and suse classes in ncf tests  

Fdall:
https://issues.rudder.io/issues/15110


ncharles:
why not defining a sles class in the test template ?

Fdall:
Commit modified

Fdall:
Commit modified

Fdall:
PR rebased

PR #993, opened 2019-06-27
Fixes #15135: Remove all write api of ncf and adapt editor to use only Rudder API

VinceMacBuche:
https://issues.rudder.io/issues/15135


VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

builder/js/ncf.js
```diff
@@ -126,11 +126,12 @@ app.directive('constraint', function($http, $q, $timeout) {
       var timeoutStatus = false;
       var timeout = $q.defer();
 
-      var request = $http.post("/ncf/api/check/parameter",data, { 'timeout' : timeout.promise }).then(
+      var request = $http.post("/rudder/secure/api/ncf/parameter/check",data, { 'timeout' : timeout.promise }).then(
           function(successResult) {
             scope.parameter.$errors= [];
-            if (! successResult.data.result) {
-              scope.parameter.$errors = successResult.data.errors;
+            console.log(successResult.data)
```
fanf: remaining `console.log`

builder/js/ncf.js
```diff
@@ -1209,8 +1209,7 @@ $scope.onImportFileChange = function (fileEl) {
 
   // Delete a technique
   $scope.deleteTechnique = function() {
-    var data = {params: {path: $scope.path}};
-    $http.delete("/ncf/api/techniques/"+$scope.selectedTechnique.bundle_name, data).
+    $http.delete("/ncf/api/techniques/"+$scope.selectedTechnique.bundle_name).
```
fanf: shouldn't this be ported to rudder API too?

builder/js/ncf.js
VinceMacBuche: oops ;)

builder/js/ncf.js
VinceMacBuche: Totally ! when i passed on this I thought i would do it in another PR, and it will give us a lot more possibilities, like checking if there is directives applied to this technique, and prevent deletion of the technique, or provide an option to force deletion and also delete Directives and Techniques ...,

fanf: OK, will do latter. Please open a ticket to track that change.

VinceMacBuche:
Commit modified

PR #996, opened 2019-07-03
Fixes #15149: Parent ticket breaks reporting when trying to use latest version and absent state  

amousset:
https://issues.rudder.io/issues/15149


amousset:
OK, merging this PR

PR #1006, opened 2019-07-09
Fixes #15126: Add verbosity level in the rudder logger

Fdall:
https://issues.rudder.io/issues/15126


tree/20_cfe_basics/log_rudder.cf
```diff
@@ -15,10 +15,11 @@
...
 #
 #####################################################################################
-
-# @name Log for Rudder
-# @description Logging output for Rudder reports
 #
+# @name Log for Rudder with mode
+# @description Logging different verbosity outputs for Rudder reports
+#
+# @parameter mode                 The mode use to make the report (can be info, normal, debug)
```
ncharles: normal is a weird verbosity level

tree/30_generic_methods/_log_v3_mode.cf
```diff
@@ -0,0 +1,46 @@
...
+#
+# @class_prefix
+# @class_parameter old_class_prefix
+
+bundle agent _log_v3_mode(mode, message, class_parameter, old_class_prefix, class_prefix, args)
+{
+  methods:
+      # Report using the appropriate bundle(s)
+      "wrapper for ${configuration.enabled_loggers}"
+        usebundle => log_rudder_mode("${mode}", "${message}", "${class_parameter}", "${old_class_prefix}", "${class_prefix}", "@{args}"),
```
ncharles: in 5.1, ncf cannot be standalone anymore.
do we *really* need this method ? can't we simply use the log_rudder_mode ?
or do you expect that we need some layer of indirection for the future ?

Fdall: You are definitely right, we do not need another layer of indirection, this will only make things less understandable.

Fdall:
retarget to 5.1, too much diff to do, I will recreate a pull request starting in 5.1

Fdall:
new PR here: https://github.com/Normation/ncf/pull/1048

PR #1028, opened 2019-07-25
Fixes #15312: broken command execution tests in 5.1

Fdall:
https://issues.rudder.io/issues/15312


amousset:
This time let's wait for the test result. But we should fedinitely modify quality-assistant to check test status.

PR #1043, opened 2019-08-19
Fixes #15507: Remove remaining function about delete api

VinceMacBuche:
https://issues.rudder.io/issues/15507


VinceMacBuche:
There is no hooks anymore, they were only used as internal rudder purpose and we decided to not port it to Rudder code

ncharles:
So you need to remove the mention of them on the README.md

VinceMacBuche:
Commit modified

VinceMacBuche:
yes, that what i did

VinceMacBuche:
Commit modified

PR #1048, opened 2019-08-21
Fixes #15126: Add verbosity level in the rudder logger

Fdall:
https://issues.rudder.io/issues/15126

### DIFF
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -16,6 +16,58 @@
 #
 #####################################################################################
 
+# @name Info log for Rudder
+# @description Logging info output for Rudder reports
+#
+# The mode parameter can be: log_repaired, log_warn, log_info, log_debug or log_trace
+#
+# @parameter mode              The log mode we want to report in
+# @parameter message           The common part of the message to display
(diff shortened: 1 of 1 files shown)


Fdall:
Commit modified

Fdall:
Commit modified

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -16,9 +16,19 @@
...
+# @description Logging different verbosity outputs for Rudder reports
+# Supported mode are:
+#  log_repaired
+#  log_warn
+#  log_info
+#  log_debug
+#  log_trace
 #
+# If the parameter mode is not in the list above, it will report in a classical manner
+# based on the reporting context and the defined classes.
```
ncharles: I'm really confused on how it should work: it has a class prefix, that it won't use to get the _repaired, _success, _failure, so the message will also be not be customized

why not simply having a 
bundle agent log_rudder_mode(mode, message, class_parameter, class_prefix) ?
signature will only expose what's relevant, and make it easier to use.

Fdall:
Commit modified

Fdall:
Commit modified

PR #1049, opened 2019-08-23
Fixes #15572: Reporting context is not re-evaluated when using multiple directives of the same technique

ncharles:
https://issues.rudder.io/issues/15572


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1062, opened 2019-09-20
Fixes #15764: Precise the documentation of user_present about user primary group

Fdall:
https://issues.rudder.io/issues/15764


Fdall:
Commit modified

PR #1074, opened 2019-09-30
Fixes #15804: missing usermod path in cfengine

Fdall:
https://issues.rudder.io/issues/15804


tree/20_cfe_basics/cfengine/paths.cf
```diff
@@ -478,11 +479,11 @@ bundle common paths
       "path[userdel]"             string => "/usr/sbin/userdel";
       "path[usermod]"             string => "/usr/sbin/usermod";
 
-    archlinux||darwin::
```
amousset: We shouldn't diverge from CFEngine, is it fixed upstream?

PR #1079, opened 2019-10-14
Fixes #13740: Service having both an init script and a real systemd unit are managed through the init script

amousset:
https://issues.rudder.io/issues/13740


amousset:
After testing it appears debian service correctly use system even when calling then from /etc/init.d (tested on debian 8), and this was very likely only necessary for Rudder services themselves.

We could maybe completely remove the exception now.

amousset:
PR updated with a new commit

amousset:
OK, squash merging this PR

PR #1088, opened 2019-10-23
Fixes #16037: Drag'N Drop not working for generic method

RaphaelGauthier:
https://issues.rudder.io/issues/16037

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -381,10 +381,10 @@
                   </div>
 
                   <div class="col">
-                    <div id="no-methods" ng-if="selectedTechnique.method_calls.length === 0" ng-click="toggleDisplay(false)">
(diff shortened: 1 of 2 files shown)


builder/index.html
```diff
@@ -381,10 +381,10 @@ <h5>
                   </div>
 
                   <div class="col">
+                    <ul id="methods" class="list-unstyled" dnd-list="selectedTechnique.method_calls" dnd-drop="dropCallback(item, index, type);" >
```
VinceMacBuche: is it working if you drag over the notification ? Shoud the notification should not be the first li element of the ul ?

RaphaelGauthier: It's working well but it would be more w3c compliant if this div was a <li> element, I will change that!

RaphaelGauthier:
Commit modified

PR #1092, opened 2019-10-29
Fixes #16078: Applying directives using the exact same GM with same parameters will always result in missing reports

Fdall:
https://issues.rudder.io/issues/16078

### DIFF
--- a/tree/30_generic_methods/directory_present.cf
+++ b/tree/30_generic_methods/directory_present.cf
@@ -27,17 +27,24 @@
 bundle agent directory_present(target)
 {
   vars:
-      "old_class_prefix"  string => canonify("directory_present_${target}");
-      "args"               slist => { "${target}" };
-      "report_param"      string => join("_", args);
-      "full_class_prefix" string => canonify("directory_present_${report_param}");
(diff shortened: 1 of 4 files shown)


tree/30_generic_methods/directory_present.cf
```diff
@@ -27,17 +27,24 @@
 bundle agent directory_present(target)
 {
   vars:
-      "old_class_prefix"  string => canonify("directory_present_${target}");
-      "args"               slist => { "${target}" };
-      "report_param"      string => join("_", args);
-      "full_class_prefix" string => canonify("directory_present_${report_param}");
-      "class_prefix"      string => string_head("${full_class_prefix}", "1000");
+      "old_class_prefix"       string => canonify("directory_present_${target}");
+      "uniq_old_class_prefix"  string => canonify("${report_data.directive_id}_directory_present_${target}");
```
ncharles: why do we need the uniq_old_class_prefix ?

Fdall:
This not good enough, we should wait to have a proper id for each call done by the generation before trying to fix this, or it will seem super hacky

PR #1108, opened 2019-11-13
Fixes #16181: Missing edit_lines promises in dry-run lib

amousset:
https://issues.rudder.io/issues/16181


ncharles:
why are you removing the user_dry_run, plus all the *_dry_run ?

amousset:
These classes were mainly targetting ncf users outside of Rudder, we never actually used them. I think we have no need for them and it makes the code more readable.

amousset:
PR updated with a new commit

amousset:
@ncharles updated

PR #1124, opened 2019-12-06
Fixes #16135: We can save a technique with empty name when we add a resource

VinceMacBuche:
https://issues.rudder.io/issues/16135


VinceMacBuche:
Commit modified

PR #1125, opened 2019-12-06
Fixes #16019: Close button in Generic Method tab is not working

VinceMacBuche:
https://issues.rudder.io/issues/16019

### DIFF
--- a/builder/index.html
+++ b/builder/index.html
@@ -156,7 +156,7 @@
           <div class="col-xl-4 col-5 col-techniques" ng-click="toggleDisplay(true)">
             <div class="row flex-column">
               <div class="col-auto">
-                <div class="row" ng-if="ui.showTechniques">
+                <div class="row">
                   <div class="col">
                     <h3 class="title">Techniques</h3>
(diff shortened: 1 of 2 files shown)


builder/index.html
```diff
@@ -156,7 +156,7 @@ <h1>You are not authenticated in ncf-builder</h1>
           <div class="col-xl-4 col-5 col-techniques" ng-click="toggleDisplay(true)">
             <div class="row flex-column">
               <div class="col-auto">
-                <div class="row" ng-if="ui.showTechniques">
```
RaphaelGauthier: This was used to prevent the user from clicking on the "Import" button when the left menu is folded. We shouldn't remove it, or find another workaround for this bug

VinceMacBuche:
Commit modified

builder/index.html
VinceMacBuche: I fixed by adding a css pointer-events, that disable click events https://css-tricks.com/almanac/properties/p/pointer-events/

PR #1126, opened 2019-12-09
Fixes #15558: "Technique diverges" popup is broken

RaphaelGauthier:
https://issues.rudder.io/issues/15558

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -57,6 +57,7 @@ app.directive('techniquename', function($filter) {
     require: 'ngModel',
     link: function(scope, elm, attrs, ctrl) {
       ctrl.$validators.techniqueName = function(modelValue, viewValue) {
+        if(viewValue===undefined) return false;
          // Get all techniqueNames in lowercase
          var techniqueNames = scope.techniques.map(function (technique,index) { return technique.name.toLowerCase()})
          // Remove the original name from the technique names array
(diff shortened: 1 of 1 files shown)


builder/js/ncf.js
```diff
@@ -939,12 +940,12 @@ $scope.onImportFileChange = function (fileEl) {
...
   // Check if a technique has not been changed, and if we can use reset function
   $scope.isUnchanged = function(technique) {
-    return (!$scope.suppressFlag && angular.equals(technique, $scope.originalTechnique));
+    return angular.equals(technique, $scope.originalTechnique);
   };
 
   // Check if a technique has been saved,
   $scope.isNotSaved = function() {
-    return $scope.originalTechnique !== undefined && $scope.originalTechnique.bundle_name === undefined;
+    return $scope.originalTechnique != undefined && $scope.originalTechnique.bundle_name === undefined;
```
VinceMacBuche: Why removing the second = in !==

builder/js/ncf.js
```diff
@@ -1133,12 +1134,13 @@ $scope.onImportFileChange = function (fileEl) {
   // selected technique is correct if:
   // * There is at least one method call
   $scope.checkSelectedTechnique = function() {
+    if($scope.selectedTechnique===undefined || $scope.selectedTechnique===undefined) return false;
```
VinceMacBuche: You check the same value twice, remove second $scope.selectedTechnique===undefined

builder/js/ncf.js
RaphaelGauthier: We do not only want to check if is `$scope.originalTechnique` is strictly different from undefined, we want to check if it is different from **null** too. I could write `($scope.originalTechnique !== undefined && $scope.originalTechnique !== undefined)` instead

builder/js/ncf.js
RaphaelGauthier: My bad, I would probably want to check if it is striclty different from null too

RaphaelGauthier:
Commit modified

RaphaelGauthier:
PR replaced by https://github.com/Normation/ncf/pull/1130

PR #1127, opened 2019-12-09
Fixes #16330: Make Result condition fields more readable

RaphaelGauthier:
https://issues.rudder.io/issues/16330


RaphaelGauthier:
It looks like this now :

result-conditions

PR #1128, opened 2019-12-10
Fixes #16101: Technique editor truncate the text in the generic method, even if the space is available to print it

RaphaelGauthier:
https://issues.rudder.io/issues/16101


RaphaelGauthier:
Before approving this, we should add a maximum height to the textareas so their content do not take too much space

RaphaelGauthier:
Commit modified

RaphaelGauthier:
Done, max-height set to 150px

PR #1133, opened 2020-01-13
Fixes #16534: yum package module is not compatible with systems having only python2

amousset:
https://issues.rudder.io/issues/16534


amousset:
OK, merging this PR

amousset:
OK, merging this PR

PR #1137, opened 2020-01-22
Fixes #16599: Basic support for categories in technique editor

VinceMacBuche:
https://issues.rudder.io/issues/16599

### DIFF
--- a/tools/ncf.py
+++ b/tools/ncf.py
@@ -112,16 +112,17 @@ def get_all_techniques_filenames(migrate_technique = False):
   if migrate_technique:
     path = os.path.join(basePath,"ncf/50_techniques")
   else:
-    path = os.path.join(basePath,"techniques/ncf_techniques")
+    path = os.path.join(basePath,"techniques")
 
   return get_all_cf_filenames_under_dir(path, not migrate_technique)
 
-
+excluded_dirs = [ "applications", "fileConfiguration", "fileDistribution", "jobScheduling", "systemSettings", "system" ]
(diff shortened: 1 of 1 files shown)


tools/ncf.py
```diff
@@ -428,8 +429,10 @@ def get_all_techniques_metadata(include_methods_calls = True, migrate_technique
     with codecs.open(file, encoding="utf-8") as fd:
       content = fd.read()
     try:
+      category = os.path.basename(os.path.dirname(os.path.dirname(os.path.dirname(file))))
```
fanf: could you comment: "we need to go up 3 directories: technique version, name, and finaly category"

tools/ncf.py
```diff
@@ -112,16 +112,17 @@ def get_all_techniques_filenames(migrate_technique = False):
   if migrate_technique:
     path = os.path.join(basePath,"ncf/50_techniques")
   else:
-    path = os.path.join(basePath,"techniques/ncf_techniques")
+    path = os.path.join(basePath,"techniques")
 
   return get_all_cf_filenames_under_dir(path, not migrate_technique)
 
-
+excluded_dirs = [ "applications", "fileConfiguration", "fileDistribution", "jobScheduling", "systemSettings" ]
```
fanf: why not `system` too?

VinceMacBuche: because i forgot system, so i guess it's a really good catch, but i prefer to say that I did it on purpose to test you if you really read my code ...

Thank you :)

VinceMacBuche:
Commit modified

PR #1139, opened 2020-01-29
Fixes #16645: Command parameter should not be called "Command name"

amousset:
https://issues.rudder.io/issues/16645

### DIFF
--- a/tree/30_generic_methods/command_execution.cf
+++ b/tree/30_generic_methods/command_execution.cf
@@ -20,27 +20,27 @@
 # @description Execute a command
 # @action
 #
-# @parameter command_name  Command name
+# @parameter command  Command to run
 # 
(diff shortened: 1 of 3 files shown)


tree/30_generic_methods/command_execution.cf
```diff
@@ -20,7 +20,7 @@
 # @description Execute a command
 # @action
 #
-# @parameter command_name  Command name
+# @parameter command_name  Command to run
```
VinceMacBuche: paremeter will still be named "Command name"  if you keep command_name

VinceMacBuche: in fact all parameter names are 

replace "_" by " " and uppercase of first character

amousset:
PR updated with a new commit

PR #1162, opened 2020-04-01
Fixes #17030: technique editor does not prevent user from cloning a technique with an already existing id

VinceMacBuche:
https://issues.rudder.io/issues/17030


VinceMacBuche:
Commit modified

PR #1171, opened 2020-04-14
Fixes #17113: Copy_from bundles are no longer keeping perms by default

Fdall:
https://issues.rudder.io/issues/17113


amousset:
There are probably similar bodies in 5.0 techniques

amousset:
We also need to check in ncf ncf as we moved bodies from techniques to ncf

Fdall:
We still need to double check the remote_copy current behaviour before merging

ncharles:
this looks much saner, but I'm pretty sure it will change a lot of behaviour. Notably when we are copying from server and then changing permissions. We'll always have repaired in this case

PR #1188, opened 2020-04-27
Fixes #17256: Parent broke policies

amousset:
https://issues.rudder.io/issues/17256

### DIFF
--- a/tree/20_cfe_basics/packages.cf
+++ b/tree/20_cfe_basics/packages.cf
@@ -24,6 +24,14 @@ body package_module test
       #default_options =>  {};
 }
 
+# Not upstream
+body package_module zypper_pattern
+{
+      query_installed_ifelapsed => "$(ncf_def.package_module_query_installed_ifelapsed)";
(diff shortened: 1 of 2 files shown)


tree/20_cfe_basics/packages.cf
```diff
@@ -24,6 +24,14 @@ body package_module test
       #default_options =>  {};
 }
 
+# Not upstream
+body package_module zypper_pattern
+{
+      query_installed_ifelapsed => "$(ncf_def.package_module_query_installed_ifelapsed)";
+      query_updates_ifelapsed => "$(ncf_def.package_module_query_updates_ifelapsed)";
+      #default_options =>  {};
```
VinceMacBuche, line 32: why it is prefixed by a dash ?

amousset: It's a comment but Github doesn't know about CFEngine syntax.

amousset: (all package modules have this field commented, I wanted to stay consistent)

PR #1194, opened 2020-05-07
Fixes #17354: remove non-rudder logger from ncf

ncharles:
https://issues.rudder.io/issues/17354


ncharles:
Commit modified

amousset:
keeping the _log_v3 in generic methods would ease upmerge

ncharles:
we'll always have upmerge issue, won't we? isn't it better to have it now before LTS than after?

amousset:
we can keep the `log_v3` name (until rudder-lang generated code)

ncharles:
Commit modified

ncharles:
Commit modified

PR #1197, opened 2020-05-18
Fixes #17095: Allow forcing a "None" component in reporting

amousset:
https://issues.rudder.io/issues/17095

### DIFF
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -37,6 +37,8 @@ bundle agent log_rudder_mode(mode, message, class_parameter, class_prefix)
       "c_class_prefix"         string => "";
       "c_class_prefix"         string => canonify("${class_prefix}");
       "component_key"          string => "${class_parameter}";
+    rudder_report_force_component_none::
+      "component_key"          string => "None";
 
     pass1.!known_mode::
       "effective_mode"         string => "log_warn";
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/log_rudder.cf
```diff
@@ -728,3 +733,27 @@ bundle agent endExecution
 * Any existing configuration policy will continue to be applied without change. *
 *********************************************************************************";
 }
+bundle agent rudder_set_force_component_none(mode)
```
ncharles, line 736: you should advertise that it is necessary to call it twice: once to set the mode, and once to remove the mode, otherwise all reports would be Nonde

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -728,3 +733,27 @@ bundle agent endExecution
...
+{
+  vars:
+    # this is a way to remove the class without side effect
+    pass2.unset_mode::
+      "ignored" string        => "",
+                      classes => if_ok_cancel("rudder_report_force_component_none");
+
+  classes:
+      # double negation, so that the default is dry_run (in case of error)
```
ncharles, line 745: the comment is not correct

tree/20_cfe_basics/log_rudder.cf
ncharles: also, that calling it with any text but "false" would make futur reports None

amousset:
PR updated with a new commit

PR #1211, opened 2020-05-29
Fixes #17577: Rework the documentation of the command_ and condition_ generic methods

Fdall:
https://issues.rudder.io/issues/17577

### DIFF
--- a/tree/30_generic_methods/command_execution.cf
+++ b/tree/30_generic_methods/command_execution.cf
@@ -19,9 +19,19 @@
 # @name Command execution
 # @description Execute a command
 # @action
+# @documentation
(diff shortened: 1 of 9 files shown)


tree/30_generic_methods/command_execution.cf
```diff
@@ -19,6 +19,16 @@
 # @name Command execution
 # @description Execute a command
 # @action
+# @documentation
+#
+# Execute the "**Command**" in shell.
```
amousset, line 24: I think bold is enough, no need for quotes

Fdall:
PR updated with a new commit

PR #1214, opened 2020-06-07
Fixes #17654: Still some technique diverge popup after you added a new method

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/17654


fanf:
Is this to mean that the `getMethodParameters` is not correct or that it is safer to not try to use it for now?

VinceMacBuche:
That method is unused, just some dead code removal

Fix is just to fix the way we create a parameter in a new method call


        return  {"name" : v['name'], "value" : undefined};

PR #1221, opened 2020-06-22
Fixes #17776: When removing and adding a new method, the save button is disabled

VinceMacBuche:
https://issues.rudder.io/issues/17776


VinceMacBuche:
PR updated with a new commit

PR #1223, opened 2020-06-25
Fixes #17793: Importing technique leads to a blocked editor

ElaadF:
https://issues.rudder.io/issues/17793

### DIFF
--- a/builder/js/ncf.js
+++ b/builder/js/ncf.js
@@ -445,19 +445,59 @@ function updateFileManagerConf () {
 // Transform a ncf technique into a valid UI technique
 // Add original_index to the method call, so we can track their modification on index
 // Handle classes so we split them into OS classes (the first one only) and advanced classes
-function toTechUI (technique) {
+function toTechUI (technique, version) {
   if ("method_calls" in technique) {
     var calls = technique.method_calls.map( function (method_call, method_index) {
       method_call["original_index"] = method_index;
 
       // Handle class_context
       defineMethodClassContext(method_call)
+
+      // support of version 1.0, method parameters are formatted like that : 'args':[param1,param2 ...]
+      // be careful position of the parameters in 'args' define in which method's parameters  it will be paired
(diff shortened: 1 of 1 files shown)


builder/js/ncf.js
```diff
@@ -1066,7 +1111,7 @@ $scope.onImportFileChange = function (fileEl) {
     return { "method_calls" : []
     , "name"         : ""
     , "description"  : ""
-    , "version"      : "1.0"
```
VinceMacBuche, line 1069: it's not the same version ;) should still be 1.0! here it is the version of the technique not the import version

builder/js/ncf.js
```diff
@@ -1337,7 +1382,7 @@ $scope.onImportFileChange = function (fileEl) {
       ncfTechnique = data.data.techniques.technique;
 
       // Transform back ncfTechnique to UITechnique, that will make it ok
-      var savedTechnique = toTechUI(ncfTechnique);
+      var savedTechnique = toTechUI(ncfTechnique, ncfTechnique['version']);
```
VinceMacBuche, line 1385: you can say it's 2.0 here.

ElaadF:
Commit modified

builder/js/ncf.js
```diff
@@ -1337,7 +1382,7 @@ $scope.onImportFileChange = function (fileEl) {
       ncfTechnique = data.data.techniques.technique;
 
       // Transform back ncfTechnique to UITechnique, that will make it ok
-      var savedTechnique = toTechUI(ncfTechnique);
+      var savedTechnique = toTechUI(ncfTechnique, 2.0);
```
VinceMacBuche, line 1385: should be a string

VinceMacBuche: In fact ... it does not work with an int, and you could simplify with 1 or 2

VinceMacBuche: ok I'm merging then

PR #1231, opened 2020-07-22
Fixes #17917: On debian, package with new dependencies is not upgraded to latest available version

fanf:
https://issues.rudder.io/issues/17917


fanf:
PR updated with a new commit

PR #1233, opened 2020-08-05
Fixes #18009: File in INI section method does not work as intended

Fdall:
https://issues.rudder.io/issues/18009


tree/30_generic_methods/file_line_present_in_ini_section.cf
```diff
@@ -34,32 +34,70 @@ bundle agent file_line_present_in_ini_section(file, section, line)
...
         classes       => classes_generic_two("${old_class_prefix}", "${class_prefix}");
 
   methods:
+    pass3::
       "sanitize" usebundle => _classes_sanitize("${class_prefix}");
       "sanitize" usebundle => _classes_sanitize("${old_class_prefix}");
       "report"   usebundle => _log_v3("Insert line(s) into ${file}", "${file}", "${old_class_prefix}", "${class_prefix}", @{args});
 }
+
+bundle edit_line append_block(block) {
```
ncharles, line 90: these bundle should go in the ncf lib

tree/30_generic_methods/file_line_present_in_ini_section.cf
```diff
@@ -34,32 +34,70 @@ bundle agent file_line_present_in_ini_section(file, section, line)
       "report_param"           string => join("_", args);
       "full_class_prefix"      string => canonify("file_line_present_in_ini_section_${report_param}");
       "class_prefix"           string => string_head("${full_class_prefix}", "1000");
-      "section_and_blank_line" string => "[${section}]
-";
+      "hash"                   string => hash("${report_param}", "md5");
+      "last_line"              string => execresult("/usr/bin/tail -n 1 ${file} #${hash}", "useshell");
```
ncharles, line 38: i'm not sure this path is the same every platform. Is it ?

Fdall:
PR updated with a new commit

tree/30_generic_methods/file_line_present_in_ini_section.cf
```diff
@@ -34,32 +34,60 @@ bundle agent file_line_present_in_ini_section(file, section, line)
       "report_param"           string => join("_", args);
       "full_class_prefix"      string => canonify("file_line_present_in_ini_section_${report_param}");
       "class_prefix"           string => string_head("${full_class_prefix}", "1000");
-      "section_and_blank_line" string => "[${section}]
-";
+      "hash"                   string => hash("${report_param}", "md5");
```
ncharles, line 37: this seems overkill. file and section should be enough

tree/30_generic_methods/file_line_present_in_ini_section.cf
```diff
@@ -34,32 +34,60 @@ bundle agent file_line_present_in_ini_section(file, section, line)
...
+      "last_line"              string => execresult("/usr/bin/tail -n 1 ${file} #${hash}", "useshell");
+
+    pass2::
+      "line_list" slist =>  splitstring("${line}", "${const.n}", "999999");
+      "unique_lines" slist =>  unique("line_list");
```
ncharles, line 42: i don't understand why this is deduplicated. nowhere in the method it is stated that it should behave like that, and this is really surprising for user

tree/30_generic_methods/file_line_present_in_ini_section.cf
```diff
@@ -34,32 +34,60 @@ bundle agent file_line_present_in_ini_section(file, section, line)
...
+      # limit the scope to the section, it does not exist yet
       "${file}"
-        create        => "true",
-        edit_line     => ncf_insert_block("${section_and_blank_line}"),
+        edit_line     => append_block("${lines_to_insert}"),
         edit_defaults => ncf_empty_select("false"),
+        create        => "true",
         ifvarclass    => "section_absent",
-        comment       => "Add section to file with a blank line";
+        comment       => "Add lines at the end of the file since the target section exists, is empty and is the last one in the file",
```
ncharles, line 67: comment should state "since target seection doesn't exists"

tree/30_generic_methods/file_line_present_in_ini_section.cf
Fdall: you can have 2 edit in the same sections and file in 2 directives, this is fine

tree/30_generic_methods/file_line_present_in_ini_section.cf
Fdall: This is the current behaviour, it currently only uses "insert_lines" promises so you can't have 2 times the same line in your section. And here it done manually since the section delimiters are limited in cfengine and do not work if the section is the last one in a file and is empty, the section automatically becomes the whole file.... so we need to create a block of text, clean its duplicated lines and insert it at the end without any duplication check.

Fdall:
PR updated with a new commit

tree/30_generic_methods/file_line_present_in_ini_section.cf
ncharles: it was ncf_insert_block, which is insert_type => "preserve_block", so it doesn't deduplicate

Fdall: This was an issue: at section creation it used ncf_insert_block, with preserver_block (which in fact still do a dedupolicate but with the whole block and not line per line), but afterward, when the section exist, it uses "ensure_line_in_ini_section" [https://github.com/Normation/ncf/blob/fc517e756083b0d05f44001d17f01e785b3a2f27/tree/30_generic_methods/file_line_present_in_ini_section.cf#L57](url)

Fdall: Which is defined here: [https://github.com/Normation/ncf/blob/fc517e756083b0d05f44001d17f01e785b3a2f27/tree/20_cfe_basics/files.cf#L232](url)

Fdall: And does apply a "deduplicate"

fanf:
Closing, because 5 years old. If still relevant, it will need a big update.

PR #1236, opened 2020-08-10
Fixes #18102: create a bundle specific pour NA reports

ncharles:
https://issues.rudder.io/issues/18102


ncharles:
still being tested

PR #1239, opened 2020-09-08
Fixes #18177: Method shared folder should not state that you can use md5 as hash method

VinceMacBuche:
https://issues.rudder.io/issues/18177

### DIFF
--- a/tree/30_generic_methods/file_from_shared_folder.cf
+++ b/tree/30_generic_methods/file_from_shared_folder.cf
@@ -21,9 +21,10 @@
 #
 # @parameter source      Source file (path, relative to /var/rudder/configuration-repository/shared-files)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256). Only used on Windows, ignored on Unix.
-# @parameter_constraint hash_type "select" : [ "md5", "sha1", "sha256" ]
-#
+# @parameter hash_type   Hash algorithm used to check if file is updated (sha256, sha512). Only used on Windows, ignored on Unix. default is sha256
+# # keep md5 and sha1 to support old generic methods, also add a default so you don't have to enter a value on linux, md5/sha1 should go to default
+# @parameter_constraint hash_type "select" : [ "", "sha256", "sha512", "md5", "sha1" ]
+# @parameter_constraint hash_type "allow_empty_string" : true
 # @class_prefix file_from_shared_folder
 # @class_parameter destination
 


VinceMacBuche:
PR updated with a new commit

tree/30_generic_methods/file_from_shared_folder.cf
```diff
@@ -21,9 +21,10 @@
 #
 # @parameter source      Source file (path, relative to /var/rudder/configuration-repository/shared-files)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256). Only used on Windows, ignored on Unix.
-# @parameter_constraint hash_type "select" : [ "md5", "sha1", "sha256" ]
-#
+# @parameter hash_type   Hash algorithm used to check if file is updated (sha256, sha512). Only used on Windows, ignored on Unix. default is sha256
+# # keep md5 and sha1 to support old generic methods, also add a default so you don't have to enter a value on linux, md5/sha1 should go to default
+# @parameter_constraint hash_type "select" : [ "", "sha256", "sha512", "md5", "sha1" ]
```
Fdall, line 26: This will still be displayed in the technique editor under the field as field info

tree/30_generic_methods/file_from_shared_folder.cf
```diff
@@ -21,9 +21,10 @@
 #
 # @parameter source      Source file (path, relative to /var/rudder/configuration-repository/shared-files)
 # @parameter destination Destination file (absolute path on the target node)
-# @parameter hash_type   Hash algorithm used to check if file is updated (md5, sha1, sha256). Only used on Windows, ignored on Unix.
-# @parameter_constraint hash_type "select" : [ "md5", "sha1", "sha256" ]
-#
+# @parameter hash_type   Hash algorithm used to check if file is updated (sha256, sha512). Only used on Windows, ignored on Unix. default is sha256
```
amousset, line 24: we can't have a default here because it would break windows method and we can't force windows plugin version sync easily

amousset: (already thought about adding one as it really inconvenient on Linux, should have added a comment here)

VinceMacBuche: I think we can update it, we know all users of the plugins and can make them upgrade, it's so much a PITA

tree/30_generic_methods/file_from_shared_folder.cf
VinceMacBuche: nope parameter description is limited to one line contrary to documentation

tree/30_generic_methods/file_from_shared_folder.cf
amousset: then you need to actually define the default value in the CFengine code with a defaults promise.

VinceMacBuche: we don't need it in cfengine, and it can be empty

VinceMacBuche: The value will still be defined but empty

amousset: indeed, we don't use it with CFEngine

ncharles (approved):

PR #1243, opened 2020-09-29
Fixes #18143: Zypper module does not work on python2

ncharles:
https://issues.rudder.io/issues/18143


ncharles:
I could not reproduce the issue
Yet, with this patch I could not either. Trying to force python2 and python3 did not trigger an error either

ncharles:
Finally managed to reproduce the issue, this pr fixes the issue

PR #1249, opened 2020-10-21
Fixes #18384: Create a file from remote template generic method

ncharles:
https://issues.rudder.io/issues/18384

### DIFF
--- a/tree/30_generic_methods/file_from_remote_template.cf
+++ b/tree/30_generic_methods/file_from_remote_template.cf
@@ -1,0 +1,106 @@
+#####################################################################################
+# Copyright 2020 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File from remote template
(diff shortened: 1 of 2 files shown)


ncharles:
this is not tested yet

ncharles:
Commit modified

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

tree/30_generic_methods/file_from_remote_template.cf
```diff
@@ -0,0 +1,106 @@
...
+bundle agent file_from_remote_template(source_template, destination, template_type)
+{
+  vars:
+      "old_class_prefix"  string => canonify("file_from_remote_template_${destination}");
+      "args"               slist => { "${source_template}", "${destination}", "${template_type}" };
+      "report_param"      string => join("_", args);
+      "full_class_prefix" string => canonify("file_from_template_${report_param}");
+      "class_prefix"      string => string_head("${full_class_prefix}", "1000");
+
+      "template_folder"            string => "${sys.workdir}/templates/file_from_remote_template";
```
peckpeck, line 73: there is a /var/rudder/tmp that already exists and the training tells user to use it for templates

ncharles: interesting, because  File from template with type recommends the templates folder

> It is common to use a specific folder to store those templates after copy, for example in `${sys.workdir}/templates/`.

peckpeck: /var/rudder/templates is not provided by the package, i think we should avoid multiplying things in /var/rudder

ncharles: so we should change the documentation on the others

ncharles:
PR updated with a new commit

tree/30_generic_methods/file_from_remote_template.cf
peckpeck: yes we should

ncharles:
PR updated with a new commit

peckpeck (approved):

PR #1251, opened 2020-10-27
Fixes #18433: Do not fix component key in generic methods reporting

peckpeck:
https://issues.rudder.io/issues/18433


amousset:
Ok for me, I'll let @ncharles have a look too.

ncharles (approved):

PR #1254, opened 2020-10-28
Fixes #18442: file augeas set class parameter should be path rather than lens

gpoblon:
https://issues.rudder.io/issues/18442


amousset (approved):

gpoblon:
PR updated with a new commit

PR #1256, opened 2020-10-29
Fixes #18447: Document the permissions recursive and permissions type recursive methods

Fdall:
https://issues.rudder.io/issues/18447


tree/30_generic_methods/permissions_recursive.cf
```diff
@@ -18,6 +18,10 @@
 
 # @name Permissions (recursive)
 # @description Verify if a file or directory has the right permissions recursively
+# @documentation The method ensure that all files and directories under `path` have
```
amousset, line 21:
```suggestion
# @documentation The method ensures that all files and directories under `path` have
```

ncharles (approved):

PR #1257, opened 2020-10-30
Fixes #18451: techniques in audit don't report correctly for editing values in files

ncharles:
https://issues.rudder.io/issues/18451

### DIFF
--- a/tree/20_cfe_basics/dry_run.cf
+++ b/tree/20_cfe_basics/dry_run.cf
@@ -198,6 +198,7 @@ bundle agent pop_dry_run_mode()
 #  - classes
 #  - methods
 #  - reports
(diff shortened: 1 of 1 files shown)


tree/20_cfe_basics/dry_run.cf
```diff
@@ -283,40 +283,6 @@ body action databases_action
     action_policy => "fix";
 }
 
-# in bundle edit_line
```
amousset, line 286: could you add a comment explaining why these aren't there? Could be in the comment explaining why other types are already skipped (line 191)

ncharles:
PR updated with a new commit

amousset (approved):

PR #1276, opened 2020-12-11
Fixes #18764: Regex constraint must not contain escaped chars

Fdall:
https://issues.rudder.io/issues/18764


VinceMacBuche (changes requested):
I don't think it shouln(t d be done here but we should fix the problem below

VinceMacBuche (approved):

PR #1282, opened 2021-02-16
Fixes #18911: implement gm condition from list variable match

gpoblon:
https://issues.rudder.io/issues/18911

### DIFF
--- a/tree/30_generic_methods/condition_from_list_variable_match.cf
+++ b/tree/30_generic_methods/condition_from_list_variable_match.cf
@@ -1,0 +1,76 @@
+#####################################################################################
+# Copyright 2021 Normation SAS
+#####################################################################################
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, Version 3.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


tree/30_generic_methods/condition_from_list_variable_match.cf
```diff
@@ -0,0 +1,76 @@
...
+
+      "${condition}_true"  and => {reglist("${variable_name}", "${value_match}"), "var_exists"},
+                                       scope => "namespace";
+
+      "${condition}_false"  or => {"!${condition}_true", "!var_exists"},
+                                       scope => "namespace";
+
+  methods:
+    pass3::
+      "failure" usebundle => _classes_failure("${old_class_prefix}"),
```
Fdall, line 62: Condition_from are usually not defining failure result_classes when the _false condition is defined since it can be used to ensure that a variable does not match a specific pattern.
The method should only generate a failure when the target variable is not defined or when  neither condition_true or condition_false are defined.

Fdall: Documenting the behaviour when the list is empty is also needed.

Fdall (changes requested):

Fdall:
Closing as the PR is very old. This was most likely initiated by the earliest version of the CIS plugin which now uses the properties to define conditions.

PR #1313, opened 2021-09-28
Fixes #20007: We should be able to return an arbitrary report from the technique editor

Fdall:
https://issues.rudder.io/issues/20007


Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

PR #1317, opened 2021-10-05
Fixes #20056: Remove ubuntu20 from the Jenkinsfile test since thehost is unstable atm

Fdall:
https://issues.rudder.io/issues/20056


Fdall:
Commit modified

PR #1322, opened 2021-11-03
Fixes #20227: Constants should be defined in the same bundle

Fdall:
https://issues.rudder.io/issues/20227


Fdall:
PR updated with a new commit

PR #1324, opened 2021-12-06
Fixes #20372: there is no validation on the variable name in generic method variable_from_smth

ncharles:
https://issues.rudder.io/issues/20372

### DIFF
--- a/tree/30_generic_methods/variable_dict.cf
+++ b/tree/30_generic_methods/variable_dict.cf
@@ -23,9 +23,11 @@
 # Be careful that using a global variable can lead to unpredictable content in case of multiple definition, which is implicitly the case when a technique has more than one instance (directive). 
 # Please note that only global variables are available within templates.
 #
-# @parameter variable_prefix The prefix of the variable name
-# @parameter variable_name   The variable to define, the full name will be variable_prefix.variable_name
-# @parameter value           The variable content in JSON format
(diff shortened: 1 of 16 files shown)


tree/30_generic_methods/variable_dict.cf
```diff
@@ -23,9 +23,11 @@
 # Be careful that using a global variable can lead to unpredictable content in case of multiple definition, which is implicitly the case when a technique has more than one instance (directive). 
 # Please note that only global variables are available within templates.
 #
-# @parameter variable_prefix The prefix of the variable name
-# @parameter variable_name   The variable to define, the full name will be variable_prefix.variable_name
-# @parameter value           The variable content in JSON format
+# @parameter            variable_prefix The prefix of the variable name
+# @parameter_constraint variable_prefix "regex" : "^[A-z0-9_]+$"
```
amousset, line 27: `A-z` matches non letter chars too.

amousset: We should use something like `A-Za-z` instead.

amousset: plus we need to make it possible to use a variable here, I don't remember if we have a mechanism on server part to skip constraint check in this case

ncharles:
PR updated with a new commit

PR #1335, opened 2022-02-23
Fixes #20806: Use the new logger v4 for service methods

amousset:
https://issues.rudder.io/issues/20806

### DIFF
--- a/tree/20_cfe_basics/services.cf
+++ b/tree/20_cfe_basics/services.cf
@@ -25,7 +25,7 @@
 # * restart
 # * reload (with an alias to refresh)
 # * is-active (with an alias to status)
-# * is-active-process (in this case, the "service" parameter is the regex to match againt process list)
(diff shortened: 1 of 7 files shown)


amousset:
PR updated with a new commit

tree/20_cfe_basics/services.cf
```diff
@@ -209,13 +196,13 @@ bundle agent ncf_services(service, action)
                                 strcmp("is-active-process", "${action}"),
                                 strcmp("is-enabled", "${action}")
                               };
-      "is_start_action"         expression => regcmp(".*start.*", "${action}");
+      "is_start_action"         expression => regcmp(".*start$", "${action}");
```
ncharles, line 199: it matches restart also. Shouldn't we have a case for that ?

amousset: restart also requires clearing the failt status, but I noticed that it also matches "is-started" which does not.

ncharles (approved):

PR #1342, opened 2022-03-14
Fixes #20885: All classes manipulations are ineffeccient because there are repeated 3 times

ncharles:
https://issues.rudder.io/issues/20885


amousset:
OK, merging this PR

PR #1343, opened 2022-03-14
Fixes #20128: Variable string from command fails when command contains control structures

ncharles:
https://issues.rudder.io/issues/20128


ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

amousset (approved):

PR #1348, opened 2022-03-28
Fixes #20944: Broken reporting for double nested method calls

amousset:
https://issues.rudder.io/issues/20944

### DIFF
--- a/tree/20_cfe_basics/log_rudder.cf
+++ b/tree/20_cfe_basics/log_rudder.cf
@@ -899,11 +899,12 @@ bundle agent method_id_push(method_id)
 bundle agent method_id_pop()
 {
   vars:
+    !pass1::
       "stack_len"                int => length("method_id_unique.stack");
+      "method_id"             string => nth("method_id_unique.stack", "0");
+
     # this must be evaluated exactly only once
     pass2::
-      # get first value in the stack
-      "method_id"             string => nth("method_id_unique.stack", 0);
       # compute stack length minus 1
       "new_len_float"         string => eval("${stack_len}-1", "math", "infix");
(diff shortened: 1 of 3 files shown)


tree/20_cfe_basics/log_rudder.cf
```diff
@@ -899,11 +899,11 @@ bundle agent method_id_push(method_id)
 bundle agent method_id_pop()
 {
   vars:
+    !pass1::
       "stack_len"                int => length("method_id_unique.stack");
+      "method_id"             string => nth("method_id_unique.stack", "0");
```
ncharles, line 904: I think that method_id should be only the stack is not empty
so here
not_empty_stack.!pass2::
"method_id"             string => nth("method_id_unique.stack", "0");

and 
pass1.!pass2::
"not_empty_stack" expression => isgreaterthan("${stack_len}", "0");

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

tree/30_generic_methods/service_stopped.cf
```diff
@@ -48,7 +48,6 @@ bundle agent service_stopped(name)
       "${report_data.method_id}_${name}" usebundle => call_method("ncf_services_check");
       "${report_data.method_id}_${name}" usebundle => ncf_services("${name}", "is-active");
       "${report_data.method_id}_${name}" usebundle => call_method_classes("${report_data.method_id}_ncf_services_check");
-      "${report_data.method_id}_${name}" usebundle => call_method_classes_caller;
```
ncharles, line 51: why are you removing it ?

ncharles (approved):

PR #1350, opened 2022-04-21
Fixes #21011: report_if_condition method does not support expressions in its input

Fdall:
https://issues.rudder.io/issues/21011


tree/30_generic_methods/report_if_condition.cf
```diff
@@ -50,6 +50,8 @@ bundle agent report_if_condition(report_message, condition)
                               if => "${condition}";
       "extended_message"  string => "${report_message} was incorrect",
                               if => not("${condition}");
+  classes:
+      "local_expression"  expression => "${condition}";
```
VinceMacBuche, line 54: You also need to change all if below from "condition" to "local_expression"

VinceMacBuche: From what i read in documentation this should be ok ...

VinceMacBuche (approved):

PR #1362, opened 2022-08-08
Fixes #19033: package method leads to report error when package with non-zero epoch is updated

ncharles:
https://issues.rudder.io/issues/19033


ncharles:
PR replaced by https://github.com/Normation/ncf/pull/1363

PR #1368, opened 2022-09-13
Fixes #21758: report_if is missing a canonification

peckpeck:
https://issues.rudder.io/issues/21758

### DIFF
--- a/tree/30_generic_methods/report_if_condition.cf
+++ b/tree/30_generic_methods/report_if_condition.cf
@@ -54,7 +54,7 @@ bundle agent report_if_condition(report_message, condition)
   classes:
       "pass2" expression => "pass1";
       "pass1" expression => "any";
-      "local_expression"  expression => "${condition}";
+      "local_expression"  expression => canonify("${condition}");
 
   methods:
    pass2::


VinceMacBuche (approved):

tree/30_generic_methods/report_if_condition.cf
```diff
@@ -54,7 +54,7 @@ bundle agent report_if_condition(report_message, condition)
   classes:
       "pass2" expression => "pass1";
       "pass1" expression => "any";
-      "local_expression"  expression => "${condition}";
+      "local_expression"  expression => canonify("${condition}");
```
VinceMacBuche, line 57: What if the condition is a compoisiion of condition like "x|y" does it not become x_y here ?

amousset: :+1: what is the problem this PR solves?

peckpeck: when we create conditions with generic methods we can pass a variable that contains anything and the created condition name is automatically canonified
to use report_if with this condition we only have the non canonified name of the condition that we can pass here, so it needs to be canonified

amousset: the `condition` parameter should be something that evaluates to a valid condition expression. If not it is normal to fail IMO.

peckpeck: this is not the same behaviour as the condition_from_* methods which canonify the condittion name

and when we create a condition within an iteration, we don't have the option to make the condition name a valid static string

PR #1375, opened 2022-11-16
Fixes #22088: make iterator report more independant

VinceMacBuche:


tree/20_cfe_basics/services.cf
```diff
@@ -257,33 +257,33 @@ bundle agent ncf_services(service, action)
...
     pass1.!pass2::
-        "${report_data.method_id}_${service}" usebundle => _classes_cancel("${report_data.method_id}");
+        "${report_data.method_id}_${service}" usebundle => _classes_cancel("${report_data.report_id}");
     #####
     # Actual command - for actions, non using systemd
     #####
     pass2.is_check_action.!is_process_action.method_found.action_ok::
       "${report_data.method_id}" usebundle => _classes_success("${class_prefix}");
-      "${report_data.method_id}" usebundle => _classes_success("${report_data.method_id}");
+      "${report_data.method_id}" usebundle => _classes_success("${report_data.report_id}");
```
amousset, line 266: this won't allow making a difference in methods calling several other methods.

amousset: all will use the same prefix, method_id is there for a reason

VinceMacBuche: method_id and report_id were using the same value

VinceMacBuche: "report_data.report_id"      string => canonify("${report_id}_${report_data.directive_id}");
    "report_data.method_id"      string => canonify("${report_id}_${report_data.directive_id}");

VinceMacBuche: So if you have the problem with report_id you had the problem with method_id

Hence this pr makes method id even more specific (and i keep use it as promiser because of this) as it allow reexecution of things that were not reexecuted

VinceMacBuche: Anyway i made this changed in 7.2 but i think it should be on master

tree/20_cfe_basics/log_rudder.cf
```diff
@@ -510,20 +510,25 @@ bundle agent _method_reporting_context(c_name, c_key)
...
+      "report_data.index_eval" string => eval("${report_data.index}+1");
+      "report_data.index"      int => int("${report_data.index_eval}");
+      "report_data.component_name" string => "${c_name}";
+      "report_data.component_key"  string => "${c_key}";
+      # Non-canonified version, to be used in reports
+      "report_data.report_id_r"    string => "${report_id}";
+      # Canonified version, to be used in policies (as class prefix)
+      # it needs the directive_id to enforce unicity
+      "report_data.report_id"      string => canonify("${report_id}_${report_data.directive_id}");
+      "report_data.method_id"      string => canonify("${report_id}_${report_data.directive_id}_${report_data.index}");
```
ncharles, line 523: this change looks *really* complex.
we need to clearly define what is expected in report, method, and why index is necessary like (why not use something else like an auto generated id from the webapp) 
there are many moving parts, in services, packages in 20_services, but also in system techniques.

VinceMacBuche: replaced usage a method_id to report_id (they were the SAME value)

VinceMacBuche: To use report_id effectively in reports and method_id where we need to identify the method

tree/20_cfe_basics/services.cf
amousset: method_id is modified each time we enter a sub-method (by the push/pop mechanism), report_id stays the same. What is in _method_reporting_context_v4 is only the initialization.

amousset:
I'm working on an update version (#1432)

PR #1377, opened 2022-11-22
Fixes #22122: Where there is not Python installed on a node, the paka…

ncharles:
…ges and templates jinja2 fails without telling why


ncharles:
https://issues.rudder.io/issues/22122

amousset (approved):

PR #1378, opened 2022-12-09
Fixes #20227: Constants should be defined in the same bundle

Fdall:
https://issues.rudder.io/issues/20227


Fdall:
Closing as this will not be planned before 8.0

PR #1390, opened 2023-04-21
Fixes #22677: Augeas set method does not correctly handle successive calls

amousset:
https://issues.rudder.io/issues/22677


amousset:
PR updated with a new commit

ncharles (approved):

amousset:
OK, squash merging this PR

PR #1392, opened 2023-05-02
Fixes #22722: Add a methd to force the secondary groups of a user

Fdall:
https://issues.rudder.io/issues/22722


Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

amousset (approved):

tree/30_generic_methods/user_secondary_group.cf
```diff
@@ -0,0 +1,159 @@
...
+#
+# Ensure that the user belongs in the given secondary group, if `force` is set,
+# the user will be force to only be part of the listed `groups`.
+#
+# #### Examples
+#
+# ```yaml
+# -name: bob must be in the printers group
+#  method: user_secondary_groups
+#    login: bob
```
amousset, line 33:
```suggestion
#  params:
#    login: bob
```

amousset (changes requested):
yaml syntax

Fdall:
PR rebased

tree/30_generic_methods/user_secondary_group.cf
```diff
@@ -0,0 +1,161 @@
...
+                                              )
+                                          );
+
+  users:
+    pass2::
+      "${login}"
+        policy => "present",
+        groups_secondary => { @{group_list} },
+        classes => classes_generic_two("${class_prefix}", "${report_data.method_id}"),
+        if => and( "user_exists",
```
ncharles, line 122: this should be only if should_check and purge_orhe_groups, isn't it?

Fdall: Yes, but sould_check should not be there. It was use in the early draft of the method, will remove it.

Fdall:
PR updated with a new commit

tree/30_generic_methods/user_secondary_group.cf
Fdall: fixed

tree/30_generic_methods/user_secondary_group.cf
```diff
@@ -0,0 +1,161 @@
...
+
+
+  defaults:
+      "force" string => "false", if_match_regex => "";
+
+  classes:
+      "user_exists"         expression => userexists("${login}");
+      "group_not_found"            not => groupexists("${group_list}");
+      "purge_other_groups"  expression => strcmp("${force}", "true");
+      "should_check"               and => { "user_exists", "!group_not_found" };
```
ncharles: it's never used, and it would be great to use it for error condition also

tree/30_generic_methods/user_secondary_group.cf
```diff
@@ -0,0 +1,160 @@
...
+        if => and( "user_exists",
+                   "!group_not_found",
+                   "purge_other_groups"
+                 );
+
+  methods:
+    # Without enforce, fallback to user_group
+    pass2.!pass3.!purge_other_groups::
+     "disable_reporting"       usebundle => disable_reporting;
+      "subcall_${group_list}"  usebundle =>  user_group(
```
ncharles, line 130: should be done only if user_exists and not group_not_found

tree/30_generic_methods/user_secondary_group.cf
```diff
@@ -0,0 +1,160 @@
...
+                                               "${login}",
+                                               "${group_list}"
+                                             );
+     "reenable_reporting"      usebundle => enable_reporting,
+                                      if => "should_report";
+
+    # Result class definition
+    pass3::
+      "error"    usebundle => _classes_failure("${prefixes}"),
+                        if => "${c_id}_error";
```
ncharles, line 140:
```suggestion
                        if => "${c_id}_error|!user_exists|group_not_found";
```

tree/30_generic_methods/user_secondary_group.cf
Fdall: ? i do not understand.
We want to run this only if the 'purge_other_groups' is not defined. The reporting is then delegated to the subcall which also check for group/user existence

tree/30_generic_methods/user_secondary_group.cf
Fdall: this si redondant with the definition of `{c_id}_error

tree/30_generic_methods/user_secondary_group.cf
ncharles: there's no point calling a submethod if we already know that it will fail

Fdall:
PR updated with a new commit

ncharles (approved):

Fdall:
OK, squash merging this PR

PR #1395, opened 2023-05-09
Fixes #22735: file augeas set doesn't when value have space -  double quote not properly escaped

peckpeck:
https://issues.rudder.io/issues/22735


ElaadF:
Tested :heavy_check_mark:  seems ok

ElaadF (approved):

PR #1406, opened 2023-08-24
Fixes #23347: Improve variable_string_from_command by using execresult_as_data

amousset:
https://issues.rudder.io/issues/23347

Remove the accumulated cruft that caused multiple bugs and use the "new" `execresult_as_data_method`, now we don't need compatibility with 6.2 anymore.

Note: Due to https://northerntech.atlassian.net/browse/CFE-4244 I had to re-introduce an ugly hack (a command at the end of the command) to allow our tests to pass (and avoid running into errors when running the same command in different methods).


amousset:
PR updated with a new commit

Fdall (approved):

amousset:
This does not work when the command comes from the test module, for some reason. We need to figure it out.

amousset:
PR updated with a new commit

amousset:
OK, squash merging this PR

PR #1413, opened 2023-10-25
Fixes #23652: iteration in technique editor with package method don't report correctly

ncharles:
https://issues.rudder.io/issues/23652

This change adds a variability based on the key for generic methods.
To allow multiple execution of generic method based on service command and package, we introduced a stack for evaluation & reporting, that uses generic method id + directive + rule as an id
This solves the issue of multiple evaluations, yet it does not work with reporting: a list of package to install, with the first in repair and the other in success return repair for all the package, breaking the reporting

We have a facility to unwrap the list in the generated technique

` 27 bundle agent iteration_gm_1(c_name, c_key, report_id, name, version, architecture, provider) {
 28   methods:
 29     "c3974f2e-d9e9-4527-ab5f-e0add1c52318_${report_data.directive_id}" usebundle => _method_reporting_context_v4("${c_name}","${c_key}","${report_id}");
 30     "c3974f2e-d9e9-4527-ab5f-e0add1c52318_${report_data.directive_id}" usebundle => package_present("${name}","${version}","${architecture}","${provider}");
 31 }`

which was introduced in https://issues.rudder.io/issues/20603

so, we could be insensitive to iteration if we add the key to the report
For packages, command and services, report_id and method_id are both used for defining classes & report condition (and I suspect that one in use in place of the other more than once); so adding in them the key allow unicity
This is was this change does and fixes the reporting issue

It needs to be more throughly tested, especially for the result condution of the GM, and for techniques, and services & command

### DIFF
--- a/tests/acceptance/default_ncf.cf.sub
+++ b/tests/acceptance/default_ncf.cf.sub
@@ -204,6 +204,7 @@ bundle agent apply_gm_v4(name, args, status, result_class, mode)
       "d_uuid"       string => canonify("d_${uuid}");
       "r_uuid"       string => canonify("r_${uuid}");
 
+      "c_arg0"       string => canonify("${arg0}");
   classes:
       "pass3" expression => "pass2";
       "pass2" expression => "pass1";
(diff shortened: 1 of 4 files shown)


amousset:
Tests are failing with reporting-related errors.

ncharles:
ha, test framework probably needs to be updated as well, given the following error message
Missing method_id class 58abab1b_be8d_4809_b257_9113759ea5fd_d_58abab1b_be8d_4809_b257_9113759ea5fd_not_kept

amousset:
We will take this PR for 8.1, but replace the component arg with a hash of the parameters.

fanf:
After more investigation, we need to do deeper change and create a logger_v5 to correctly handle the case. This is a too big change for 8.1, we are forced to switch it to 8.2.
Making it a draft.

amousset:
I'm working on an update version (#1432)

PR #1418, opened 2024-03-14
Fixes #20853: sysctl value generic method is not reporting

ncharles:
https://issues.rudder.io/issues/20853


ncharles:
PR updated with a new commit

Fdall (approved):

PR #1420, opened 2024-03-19
Fixes #24537: Regex in file replace line must not match the string used as replacement

ElaadF:
https://issues.rudder.io/issues/24537


ElaadF:
Commit modified

PR #1437, opened 2024-11-21
Fixes #25924: Add a new type constraint to the variable_dict generic method

Fdall:
https://issues.rudder.io/issues/25924


amousset:
it breaks a test

PR #1438, opened 2024-11-22
Fixes #25936: Port condition_from methods to log v4+

Fdall:
https://issues.rudder.io/issues/25936

### DIFF
--- a/tree/30_generic_methods/condition_from_expression.cf
+++ b/tree/30_generic_methods/condition_from_expression.cf
@@ -80,8 +80,8 @@ bundle agent condition_from_expression(condition, expression)
                                        scope => "namespace";
 
   methods:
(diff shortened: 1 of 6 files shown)


Fdall:
Commit modified

tree/30_generic_methods/condition_from_expression.cf
```diff
@@ -84,13 +80,11 @@ bundle agent condition_from_expression(condition, expression)
                                        scope => "namespace";
 
   methods:
-      "success" usebundle => _classes_success("${old_class_prefix}");
+      "success" usebundle => _classes_success("${report_data.method_id}");
```
amousset, line 83:
```suggestion
      "${report_data.method_id}" usebundle => _classes_success("${report_data.method_id}");
```

for consistency I think we should always use it is promiser for methods.

